PR #2, opened 2017-12-31
Package workspace

RyanBerge:


collinsc:
Looks good so far, two questions:
1. Does it build?
2. Where do we want to put things like automation scripts?

I have already written a few for setting up the ros distributed network on boot, and stuff like that. We can catologue them in here, or in a separate repo.

The only issue is that these scripts might be dependent on being in a specific directory, and not necessarily portable to anything but the jetson.

RyanBerge:
@collinsc  It builds.  It doesn't do much; there's a node in each package you can spin up that advertises on the topic <package_name>/Echo which just broadcasts that it's online (I do not have launch files for this, but you can launch the nodes manually).  The idea was for this to be a starting place for people to go for writing their code.

I'm not sure where scripts should go.  I have two scripts in the root directory, one for building the repo and the tests and one for running the tests.  Those scripts are for developers working on their computer, not the Jetson.

AdamuKaapan:
As far as automation scripts like Travis CI scripts, those have to go in the base of the repo I think? I have no input on where other scripts should go. I'm thinking that if they're for a specific package they go in there, or if they're for the whole project they go in the root directory.

AdamuKaapan:
@RyanBerge I cloned the repository and after a quick issue with GTest installation (now resolved), I'm encountering build issues with tfr_communication. catkin_make is saying that it attempted to try and find the library tfr_utilities but couldn't find it. CMake output here: https://pastebin.com/xr0FpfkZ

Is this just a setup issue on my end or is it possibly more related to the fact that your projects got set up incrementally while mine is all at once?

collinsc:
I wasn't really talking about that kind of automation. I was talking about automating development tasks.  For example I have a script that starts up roscore, finds it's ip address and displays it to the user so they can remote in. I'll also have a script that takes in an IP as a parameter and sets up all of the appropriate Bash variables to allow us to connect to the ROS network on the jetson.

Would it make sense to have a scripts folder for those kinds of things? They can pile up quickly on systems like this.

@AdamuKaapan @RyanBerge

RyanBerge:
@collinsc If it turns out we have a lot of scripts and decide we want different organization, we can change that later.  Right now, because the root of the repo is about the `src` directory, there's no code or generated files in the root where the scripts are now, just `.gitignore` and the `README.md` (which I should update, actually).  I think it's an okay place to have our developer scripts.  It's where they were kept where I worked (though we obviously weren't in a ROS environment); I think we had about eight of them or so, and that wasn't too cluttered.

collinsc (approved):
It's a good start, let's stick with the tfr package prefix like we talked about, and keep evolving it through further pull requests, thanks for setting it up

PR #3, opened 2018-01-04
Rtab setup

collinsc:
# Goals
- implement some basic infrastructure.
- get rtabmap up and running.
- create basic testing infrastructure to allow for sensor detection and movement in gazebo

# Works Completed
Speed was a priority here, so I adapted some old work from gazebo and implemented a robot model, called proto, proto is described in the tfr_description, and tfr_control packages. All related files are given the proto prefix.

I also created a test course which has a hallway leading into a room with a few obstaclesin the gazebo simulation, and saved it in the tfr_gazebo worlds folder.

I created the tfr_visualization package which is where we can store various different rviz visualization settings for our data.

I also created a testing configuration in the tfr_launch package called rtab_test.launch. This will set up the robot in the world and start rtab map, and an rviz view showing the pointcloud and generated map.

# Results
So what I found here is that I don't think the plugins for stereo cameras in gazebo are really going to do the thing we want. They are spotty, hard to configure, and require so much effort any time saved by not having to run on hardware is paid back in full. I think trying to make that happen for the navigation stack will be folly.

So I propose changing strategies and moving into hardware testing early. However I accomplished what I set out to do, and there is enough progress made to check my work in.


src/tfr_control/src/ProtoDriverStation.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    window.clear(sf::Color::Black);
+
+    window.display();
+    ros::spinOnce();
+    window.setFramerateLimit(30);
+  }
+
+  return 0;
+}
```
RyanBerge: Not that it matters too much, but let's keep our .cpp files all formatted the same way; that is, 4-space tabs and aligned braces.  I'm not sure how I feel about `case` statements wrapped in braces, but it's probably fine.

src/tfr_control/src/ProtoDriverStation.cpp
```diff
@@ -0,0 +1,65 @@
+/**
```
AdamuKaapan: File name should be all lowercase: should be proto_driver_station.cpp

src/tfr_description/xacro/proto_drivebase_controller.xacro
```diff
@@ -0,0 +1,49 @@
...
+      <rightFrontJoint>wheel_fr_joint</rightFrontJoint>
+      <leftRearJoint>wheel_rl_joint</leftRearJoint>
+      <rightRearJoint>wheel_rr_joint</rightRearJoint>
+      <wheelSeparation>${base_width}</wheelSeparation>
+      <wheelDiameter>${wheel_radius*2}</wheelDiameter>
+      <robotBaseFrame>god</robotBaseFrame>
+      <torque>20</torque>
+      <topicName>cmd_vel</topicName>
+      <commandTopic>cmd_vel</commandTopic>
+      <!-- we are simulating the output of sensor fusion right here -->
```
AdamuKaapan, line 40: Is this actually producing fused odometry? Or are you just saying that, in the absence of other sensors to do fusion, we're just using this as the fused odometry even though it's only one source?

AdamuKaapan: (if that's the case, no worries, I'm just clarifying)

src/tfr_description/xacro/proto_sensors.xacro
```diff
@@ -0,0 +1,52 @@
...
+        <noise>
+          <type>gaussian</type>
+          <mean>0.0</mean>
+          <stddev>0.0007</stddev>
+        </noise>
+      </camera>
+  </xacro:macro>
+  
+  <gazebo reference="camera_link">
+      <sensor type="multicamera" name="stereo_camera">
```
AdamuKaapan: Indentation issue, there should be one less tab here before <sensor>

src/tfr_description/xacro/proto_drivebase_controller.xacro
```diff
@@ -0,0 +1,49 @@
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="proto">
+  <!--xacro dectiptio of the prototype skid steer controller-->
```
AdamuKaapan: Typo: I'm not sure, but is this supposed to be "depiction"?

src/tfr_launch/package.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>tfr_launch</name>
+  <version>0.0.0</version>
+  <description>contains various system level starting configurations to accomplish various tasks</description>
```
AdamuKaapan: Capitalize "contains"?

src/tfr_visualization/package.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>tfr_visualization</name>
+  <version>0.0.0</version>
+  <description>The tfr_visualization package</description>
```
AdamuKaapan: More descriptive description would be nice so we know exactly what goes in this package? Is it just rviz scripts? Any nodes that involve visualization even if they're custom written?

AdamuKaapan (changes requested):
Left various comments regarding a couple questions and a couple typos.

src/tfr_control/src/ProtoDriverStation.cpp
collinsc: Good call, thanks for catching that we really should incorporate this style guide in future reviews: http://wiki.ros.org/CppStyleGuide

src/tfr_description/xacro/proto_drivebase_controller.xacro
collinsc: depiction, I'll fix it

src/tfr_control/src/ProtoDriverStation.cpp
AdamuKaapan: It's already in the style guide, yup. Well, at least the naming is anyways

src/tfr_description/xacro/proto_drivebase_controller.xacro
collinsc: The reason I did that was to make sure that rtab map was subscribed to the correct topic, the idea was that we would be able to swap the gazebo simulation with the real thing with no issues.

AdamuKaapan: Got it, I figured it was something like that.

src/tfr_visualization/package.xml
collinsc: I just forgot to add it, I'll elaborate.

collinsc:
@AdamuKaapan I implemented the recommended changes. Thanks for catching them.

AdamuKaapan (dismissed):

collinsc:
@AdamuKaapan @RyanBerge I had to fix a merge conflict with dependencies, it was one line so it needs review again

AdamuKaapan (dismissed):

RyanBerge (changes requested):
I added a comment on `proto_driver_station.cpp` for changing the format to match our other source files.

src/tfr_control/src/ProtoDriverStation.cpp
collinsc: I am following the ros c++ style guide. There is no guideline established for this in our coding standard, so I would prefer ros, which is 2 spaces for indents, with softtabs.

collinsc: lets' talk at the meeting

AdamuKaapan (approved):

RyanBerge (approved):

PR #4, opened 2018-01-08
Add example ActionServer and ActionClient in Utilities

RyanBerge:
- Also fix build dependency issue


collinsc:
@RyanBerge what's the dependency issue?

src/tfr_communication/package.xml
```diff
@@ -15,6 +15,7 @@
   <test_depend>gtest</test_depend>
   <depend>roscpp</depend>
   <depend>std_msgs</depend>
+  <depend>tfr_msgs</depend>
```
collinsc, line 18: I don't know if I agree with adding in dependencies before, it's explicitly called for in the package.   Usually with files dealing with the build and packages, what has worked for me in the past has been to keep them as small as possible, then slowly add things one at a time when things break. 

I think it makes programmers more conscious of the changes they are making, no one is cleaning up after you, and if forces you to be conscious about how you do the build.

@RyanBerge What was your justification for doing it?

@AdamuKaapan Whats your opinion on this?

collinsc:
What do you think about on future pull requests that don't involve unit tests, making it a standard practice to submit a launch file demonstrating your code?

src/tfr_utilities/src/example_action_server.cpp
```diff
@@ -0,0 +1,63 @@
...
+    ros::init(argc, argv, "example_server");
+    ros::NodeHandle n;
+
+    // The defined Server namespace (do_action) should be the same as the 
+    // defined Client namespace.
+    Server server(n, "do_action", boost::bind(&execute, _1, &server), false);
+    server.start();
+    ros::spin();
+    return 0;
+}
```
AdamuKaapan: Complaint from GitHub about no newline at end of file: do we care about this?

src/tfr_utilities/package.xml
```diff
@@ -14,5 +14,6 @@
   <buildtool_depend>catkin</buildtool_depend>
   <test_depend>gtest</test_depend>
   <depend>roscpp</depend>
+  <depend>tfr_msgs</depend>
 
 </package>
```
AdamuKaapan: Complaint from GitHub about no newline at end of file: do we care about this?

src/tfr_utilities/src/example_action_client.cpp
```diff
@@ -0,0 +1,68 @@
...
+ * 
+ ***************************************************************************************/
+
+#include <tfr_msgs/ExampleAction.h> // Note: "Action" is appended
+#include <actionlib/client/simple_action_client.h>
+
+typedef actionlib::SimpleActionClient<tfr_msgs::ExampleAction> Client;
+
+// Called every time feedback is received for the goal
+void Feedback(const tfr_msgs::ExampleFeedbackConstPtr& feedback)
```
AdamuKaapan: Method name convention is `lower_case`

src/tfr_utilities/src/example_action_client.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+    ROS_INFO("Got Feedback: %f", feedback->feedback);
+}
+
+// Called once when the goal completes
+void Finished(const actionlib::SimpleClientGoalState& state, const tfr_msgs::ExampleResultConstPtr& result)
```
AdamuKaapan: Method name convention is `lower_case`

AdamuKaapan (changes requested):
Added a few comments regarding end-of-file complaints from GitHub (not sure if these matter or not, but figured they're worth bringing up just in case, feel free to ignore) and naming conventions. That said, *thank you* for removing those comments from the build files, I've been considering doing that myself.

collinsc:
There are some pretty bad build errors here with messages, I am in the process of investigating on a local copy.

AdamuKaapan:
@collinsc I'll crosscheck that in a little while once I finish some of my homework.

collinsc:
I know what's wrong. It looks like the cmake and package files were edited incorrectly. See this tutorial http://docs.ros.org/hydro/api/catkin/html/howto/format2/building_msgs.html

collinsc:
I found the source of the problems it looks like there were some steps in the tutorial skipped about the cmakelists.txt, I am in the process of checking it out on my local machine.

AdamuKaapan:
Once you get them solved post them here and I'll test on my machine as well @collinsc

AdamuKaapan:
I'll go ahead and pull and test the latest commit, just a sec.

AdamuKaapan (approved):

AdamuKaapan:
Build successful on my end after a clean, did a test run and it all seems to be working as intended.

collinsc (commented):
Let me know what you guys think about those extra manifest dependencies. I'll approve when you let me know what you think

src/tfr_communication/package.xml
AdamuKaapan: I think for most cases @collinsc I agree with you. I think this is the one exception because it's one of our unified packages that's meant to be shared throughout the whole project, and I think we'll be using it so much that leaving it out will just result in longer debug times trying to find out where the build is breaking when it'll just inevitably be added in again for almost every package. For cases aside from this one though, I agree, we should keep that down to a minimum.

collinsc (approved):

PR #6, opened 2018-01-10
Setup Jenkins build server

AdamuKaapan:
The only change necessary in the repo is adding exit codes to the build script. All other changes were done in a setup in a new MS Azure virtual machine.

With this setup (which is currently set to only this branch and the `TravisSetup` branch (which will soon be deleted), it'll apply to more once this PR is merged), all commits should now have an indicator by them indicating if a build is in progress, failed, or succeeded. Both the build and the tests have to succeed for the build to succeed. If you click on the indicator icon you can see more about the build status, including the compile log. The builds are remarkably fast so you don't have to wait long after a push to see the results.

### DIFF
--- a/build_all
+++ b/build_all
@@ -15,13 +15,16 @@ if [ $? -eq 0 ]; then
 		echo ""
 		echo "--------------------------------- Build Succeeded --------------------------------"
 		echo ""
+		exit 0
 	else
 		echo ""
 		echo "------------------------------- Building Tests Failed ----------------------------"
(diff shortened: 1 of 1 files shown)


AdamuKaapan:
(disregard the fact that there's two builds on this commit, this is because I changed the name of the build server after this commit and it doesn't remove the old build. This won't be an issue in future commits)

RyanBerge (approved):

PR #7, opened 2018-01-11
Mti setup

collinsc:
# Description
I made the xsens work and rules compliant. 

There is a whole bunch of work that went into this that is not in the pull request. It took a day of pouring over manuals, contacting support, having the firmware get corrupted, fixing it with a few python scripts. Tuning the firmware parameters, and reading the driver.

It works now, and I can say without a shadow of a doubt, not a single data point from the magnetometer touches our data. 

# Statement of Work
I adapted the ros driver to fit our constraints, which can be found here:
- https://github.com/xsens/xsens_mti_ros_node
This involved repackaging and renaming their xsens_driver and custom_msgs packages. I gave them both the package prefix of xsens for consistency. 

The actual work that I did was, add covariances to the imu data. (these might need to be modified in the future to make sensor fusion work. Strip device timestamps and replace them with our roscore timestamps to make them synchronized with the rest of the system.

I did _not_ spend any time making the drivers adherent to our code standard. They have operated seamlessly thus far, and I am afraid that refactoring them to use spaces and meet our tab requirements could introduce bugs or subtle errors, because python is whitespace sensistive. I didn't think poring through thousands of lines of python code to get every last space fixed on something we never want to touch or modify again, sounded like a very productive use of my time.

The rest of the work I did cannot be reviewed because it was at the firmware level.

# Testing
This is hardware dependent, so you will just have to take my word for it.


Jebbs:
superseded by #8

PR #8, opened 2018-01-12
Mti set up, Duo3d setup, and sensor platform integration

collinsc:
# Description
## MTi
I made the xsens work and rules compliant. 

There is a whole bunch of work that went into this that is not in the pull request. It took a day of pouring over manuals, contacting support, having the firmware get corrupted, fixing it with a few python scripts. Tuning the firmware parameters, and reading the driver.

It works now, and I can say without a shadow of a doubt, not a single data point from the magnetometer touches our data. 

## Duo3d
I integrated the duo camera, and development platform, and did some sensor bug fixes. Most of the config files were built using dynamic reconfigure. Mostly this was a boatload of IT work.

# Statement of Work

## MTi
I adapted the ros driver to fit our constraints, which can be found here:
- https://github.com/xsens/xsens_mti_ros_node
This involved repackaging and renaming their xsens_driver and custom_msgs packages. I gave them both the package prefix of xsens for consistency. 

The actual work that I did was, add covariances to the imu data. (these might need to be modified in the future to make sensor fusion work. Strip device timestamps and replace them with our roscore timestamps to make them synchronized with the rest of the system.

I did _not_ spend any time making the drivers adherent to our code standard. They have operated seamlessly thus far, and I am afraid that refactoring them to use spaces and meet our tab requirements could introduce bugs or subtle errors, because python is whitespace sensistive. I didn't think poring through thousands of lines of python code to get every last space fixed on something we never want to touch or modify again, sounded like a very productive use of my time.

The rest of the work I did cannot be reviewed because it was at the firmware level.

## Duo3d
I followed the standard arm installation instructions provided by duo with a few modifications. The procedure they give for setting linux kernel options is super hacky, and does not translate well to the tx2. I ended up using the scripts provided by jetsonhacks to recompile the kernel because they are awesome. Then I needed some help from duo support to get the arm 64 bit binaries.

Once I got the proper code I needed to get it working on the x86_64 and aarch64 software architectures this involved editing our build_all script.

Finally the camera  started to work, and after some tests and fussing about, I used the setup tutorials provided by duo to get it up and running. 

The next step was to integrate the sensor platform, this involved a few things:
- Adding transforms for the duo, and mti
- Fixing mti driver data synchronization
- Creating visualization scripts to evaluate data quality, and sensor performance.
- Tuning the performance of both drivers

# Testing
## MTi
This is hardware dependent, so you will just have to take my word for it.

## Duo3d
So on the jetson we can test this with:
`roslaunch tfr_launch sensor_test.launch`

However for review, we can just evaluate the code.


Jebbs:
Note: I edited the title and description since this PR actually included work for the MTi and Duo

Jebbs (approved):
Normally I would not approve such a huge PR, but I understand that it would be a huge pain to separate it at this point and that this needs to get merged to continue work.

Otherwise, LGTM. Thanks for adding notes to the parts you changed.

PR #11, opened 2018-01-17
DrivebaseController: Add DrivebaseController

RyanBerge:
- drivebase.cpp is a node launched with drivebase.launch. It uses
  a DrivebasePublisher class to convert Twist messages to differential
  steering angular velocity values for the two motor sets.

- Add unit tests for twistToDifferential() function

- Regenerated top-level CMakeLists.txt file --- I couldn't build the
  project with the CMakeLists file that is on master; deleting it and
  regenerating it with catkin_make fixed the problem.

- Reorganized tfr_control/CMakeLists.txt to make it easier to read

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,1 +1,63 @@
-/opt/ros/kinetic/share/catkin/cmake/toplevel.cmake
+# toplevel CMakeLists.txt for a catkin workspace
+# catkin/cmake/toplevel.cmake
+
+cmake_minimum_required(VERSION 2.8.3)
+
+set(CATKIN_TOPLEVEL TRUE)
+
+# search for catkin within the workspace
+set(_cmd "catkin_find_pkg" "catkin" "${CMAKE_SOURCE_DIR}")
+execute_process(COMMAND ${_cmd}
+  RESULT_VARIABLE _res
+  OUTPUT_VARIABLE _out
+  ERROR_VARIABLE _err
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  ERROR_STRIP_TRAILING_WHITESPACE
+)
+if(NOT _res EQUAL 0 AND NOT _res EQUAL 2)
+  # searching fot catkin resulted in an error
+  string(REPLACE ";" " " _cmd_str "${_cmd}")
+  message(FATAL_ERROR "Search for 'catkin' in workspace failed (${_cmd_str}): ${_err}")
+endif()
+
+# include catkin from workspace or via find_package()
+if(_res EQUAL 0)
+  set(catkin_EXTRAS_DIR "${CMAKE_SOURCE_DIR}/${_out}/cmake")
+  # include all.cmake without add_subdirectory to let it operate in same scope
+  include(${catkin_EXTRAS_DIR}/all.cmake NO_POLICY_SCOPE)
+  add_subdirectory("${_out}")
+
+else()
+  # use either CMAKE_PREFIX_PATH explicitly passed to CMake as a command line argument
+  # or CMAKE_PREFIX_PATH from the environment
+  if(NOT DEFINED CMAKE_PREFIX_PATH)
+    if(NOT "$ENV{CMAKE_PREFIX_PATH}" STREQUAL "")
+      string(REPLACE ":" ";" CMAKE_PREFIX_PATH $ENV{CMAKE_PREFIX_PATH})
+    endif()
+  endif()
+
+  # list of catkin workspaces
+  set(catkin_search_path "")
+  foreach(path ${CMAKE_PREFIX_PATH})
+    if(EXISTS "${path}/.catkin")
+      list(FIND catkin_search_path ${path} _index)
+      if(_index EQUAL -1)
+        list(APPEND catkin_search_path ${path})
+      endif()
+    endif()
+  endforeach()
+
+  # search for catkin in all workspaces
+  set(CATKIN_TOPLEVEL_FIND_PACKAGE TRUE)
+  find_package(catkin QUIET
+    NO_POLICY_SCOPE
+    PATHS ${catkin_search_path}
+    NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+  unset(CATKIN_TOPLEVEL_FIND_PACKAGE)
+
+  if(NOT catkin_FOUND)
+    message(FATAL_ERROR "find_package(catkin) failed. catkin was neither found in the workspace nor in the CMAKE_PREFIX_PATH. One reason may be that no ROS setup.sh was sourced before.")
+  endif()
+endif()
+
+catkin_workspace()
--- a/src/tfr_control/include/tfr_control/drivebase_publisher.h
+++ b/src/tfr_control/include/tfr_control/drivebase_publisher.h
@@ -1,0 +1,57 @@
+/****************************************************************************************
+ * File:            drivebase_publisher.h
+ * 
+ * Purpose:         This class is used by the drivebase node to convert from a Twist
+ *                  message in /cmd_vel to angular velocity for the two motor sets
+ *                  (left and right), handling the calculations for differential steering.
+ * 
+ *                  Every message read from /cmd_vel will result in two messages sent,
+ *                  one to each motor controller. This class will only publish after
+ *                  reading a message on /cmd_vel.
+ * 
+ * Subscribed To:   /cmd_vel
+ * Publishes To:    /left_tread_velocity_controller/command
+ *                  /right_tread_velocity_controller/command
+ ***************************************************************************************/
+#ifndef DRIVEBASE_PUBLISHER_H
+#define DRIVEBASE_PUBLISHER_H
+
+#include "ros/ros.h"
+#include "geometry_msgs/Twist.h"
+
+namespace tfr_control
+{
+    class DrivebasePublisher
+    {
+    public:
+        DrivebasePublisher() = delete;
+        explicit DrivebasePublisher(ros::NodeHandle& n, float axle_length, float wheel_radius);
+        DrivebasePublisher(const DrivebasePublisher& other) = delete;
+        DrivebasePublisher(DrivebasePublisher&&) = delete;
+
+        ~DrivebasePublisher() = default;
+
+        DrivebasePublisher& operator=(const DrivebasePublisher&) = delete;
+        DrivebasePublisher& operator=(DrivebasePublisher&&) = delete;
+        
+
+        // Making this static allows for unit-testing without 
+        // creating a DrivebasePublisher object
+        // left_tread and right_tread are output parameters; the rest are inputs.
+        static void twistToDifferential(float linear_v, float angular_v, float wheel_radius, 
+                                        float axle_length, float& left_tread, float& right_tread);
+
+    private:
+        void subscriptionCallback(const geometry_msgs::Twist::ConstPtr& msg);
+
+        ros::NodeHandle& n;
+        float wheel_radius;
+        float axle_length;
+
+        ros::Publisher left_tread_publisher;
+        ros::Publisher right_tread_publisher;
+        ros::Subscriber subscriber;
+    };
+}
+
+#endif // DRIVEBASE_PUBLISHER_H
--- a/src/tfr_control/include/tfr_control/drivebase_publisher.h
+++ b/src/tfr_control/include/tfr_control/drivebase_publisher.h
@@ -25,7 +25,7 @@ namespace tfr_control
     {
     public:
         DrivebasePublisher() = delete;
-        explicit DrivebasePublisher(ros::NodeHandle& n, float axle_length, float wheel_radius);
+        explicit DrivebasePublisher(ros::NodeHandle& n, float wheel_span, float wheel_radius);
         DrivebasePublisher(const DrivebasePublisher& other) = delete;
         DrivebasePublisher(DrivebasePublisher&&) = delete;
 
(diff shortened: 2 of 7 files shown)


collinsc:
Hey administration is tagged does that mean I am supposed to review this?

RyanBerge:
@collinsc  Yeah, maybe.  The original intent was to have ~2 reviewers for each pull request, where one is the person from the circle and one is "someone from administration", but since administration has only three people in it and I'm one and Adam is the other, that leaves just you.  Not quite ideal, but I do think I want two reviewers on each pull request.

collinsc:
Oh geez me and jeremy haven't been following that. My understanding was you were reviewed by whoever pointed to you in the circle. However we never talked about having an administration team as a team. So maybe let's bring that up at the meeting.

I was on the group because I needed to change a setting, you were on the group because you started the repository, and adam was on there because he is the technical director. 

If we end up having an administration team let's make it intentional.

collinsc:
In the mean time do you just want me to review it while we sort that?

RyanBerge:
It probably doesn't need to be a hard rule.  We'll talk about it at the next meeting; I just generally prefer having a couple reviewers.  @AdamuKaapan is my official reviewer so you don't need to feel pressured to get this done if you're busy.

src/CMakeLists.txt
```diff
@@ -1 +1,63 @@
-/opt/ros/kinetic/share/catkin/cmake/toplevel.cmake
\ No newline at end of file
+# toplevel CMakeLists.txt for a catkin workspace
```
collinsc, line 1: Why did the top level cmake change? that's a generated file at repository setup.

collinsc: Not a required change, just curious

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,49 @@
...
+#include "ros/ros.h"
+#include "geometry_msgs/Twist.h"
+
+namespace tfr_control
+{
+    class DrivebasePublisher
+    {
+    public:
+        DrivebasePublisher(ros::NodeHandle& n, float axle_length, float wheel_radius);
+        DrivebasePublisher(const DrivebasePublisher& other) = delete;
```
collinsc: Explicitly deleting things is good, but if we are going to do it you should delete the other generated methods.

This is the stuff that would get generated by a c++11 compiler
class Thing {
public:
  //not generated in your case Thing();                        // default constructor
    Thing(const Thing&);            // copy c'tor
    Thing& operator=(const Thing&); // copy-assign
//your discretion    ~Thing();                       // d'tor
    // C++11:
    Thing(Thing&&);                 // move c'tor
    Thing& operator=(Thing&&);      // move-assign
};

src/tfr_control/test/test_tfr_control.cpp
```diff
@@ -4,6 +4,54 @@
  * This is the main test class for the tfr_control package.
  */
 #include <gtest/gtest.h>
+#include "drivebase_publisher.h"
+
+using tfr_control::DrivebasePublisher;
```
collinsc, line 9: these are good tests,  I like the expect near macro

collinsc (changes requested):
If we are going to explicitly delete stuff, we should do it all the way. See my note about generated methods in c++11.

src/tfr_control/include/tfr_control/drivebase_publisher.h
collinsc: good article on generated methods in c++ 11
https://stackoverflow.com/questions/3734247/what-are-all-the-member-functions-created-by-compiler-for-a-class-does-that-hap

good article on asking whether you would want to delete a destructor
https://stackoverflow.com/questions/9091957/explicitly-deleting-destructors-and-not-calling-delete

src/CMakeLists.txt
RyanBerge: It doesn't build on my computer when it's just that path.  The file at that path exists, but my catkin couldn't parse the CMakeLists for some reason.

src/tfr_control/include/tfr_control/drivebase_publisher.h
RyanBerge: Good idea. Mostly I just wanted to give a hint to someone down the line who might invoke the copy constructor that we probably don't want to have multiple copies of this object, since the constructor interacts with the ROS infrastructure, subscribing to topics and stuff like that.

src/CMakeLists.txt
RyanBerge: Everyone else I've asked can build it just fine, including the Jenkins server.  I'm not sure why my system is different.

RyanBerge:
@collinsc @AdamuKaapan I just pushed changes adding in explicit `delete` and `default` keywords on the generated member functions for `drivebase_publisher.h`.  I also figured out how to get git to give a new timestamp to a commit when using `git commit --amend`, but I just realized that using the `--amend` flag during pull requests makes it harder for the reviewers to see what I changed after requesting feedback.  I'll avoid that in the future (and just create new commits for each round of feedback).

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,57 @@
...
+#include "ros/ros.h"
+#include "geometry_msgs/Twist.h"
+
+namespace tfr_control
+{
+    class DrivebasePublisher
+    {
+    public:
+        DrivebasePublisher() = delete;
+        explicit DrivebasePublisher(ros::NodeHandle& n, float axle_length, float wheel_radius);
```
AdamuKaapan: Do we want to include a definition of terms somewhere? I know for me at least "axle length" has been a confusing term (I mean, I know what it means now, but others might not: I initially thought it was the length of the axle holding the wheel on, which is like 12" long), and there's probably other confusing/ambiguous somewhere

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,57 @@
...
+        ~DrivebasePublisher() = default;
+
+        DrivebasePublisher& operator=(const DrivebasePublisher&) = delete;
+        DrivebasePublisher& operator=(DrivebasePublisher&&) = delete;
+        
+
+        // Making this static allows for unit-testing without 
+        // creating a DrivebasePublisher object
+        // left_tread and right_tread are output parameters; the rest are inputs.
+        static void twistToDifferential(float linear_v, float angular_v, float wheel_radius,
```
AdamuKaapan: Naming convention is `lower_case`

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,57 @@
...
+        
+
+        // Making this static allows for unit-testing without 
+        // creating a DrivebasePublisher object
+        // left_tread and right_tread are output parameters; the rest are inputs.
+        static void twistToDifferential(float linear_v, float angular_v, float wheel_radius, 
+                                        float axle_length, float& left_tread, float& right_tread);
+
+    private:
+        void subscriptionCallback(const geometry_msgs::Twist::ConstPtr& msg);
```
AdamuKaapan: Naming convention is `lower_case`

src/tfr_control/src/drivebase.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+
+int main(int argc, char **argv)
+{
+    ros::init(argc, argv, "drivebase");
+
+    ros::NodeHandle n;
+    float axle_length, wheel_radius;
+    n.getParam("axle_length", axle_length);
+    n.getParam("wheel_radius", wheel_radius);
```
AdamuKaapan: I think we should either add default values here (which can be done with a slightly different method: `n.param<float>("axle_length", axle_length, [some default value])`), or have a check to make sure that the values actually get loaded and is non-zero.

AdamuKaapan: Possibly add a call to ROS_WARN as well if the values aren't loaded and it uses the defaults?

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,57 @@
...
+
+        ros::NodeHandle& n;
+        float wheel_radius;
+        float axle_length;
```
AdamuKaapan: Do we want to make these `const`? That way it prevents any accidental modification once set by the constructor and makes it obvious that they're not meant to be changed/set at startup.

src/tfr_control/src/drivebase_publisher.cpp
```diff
@@ -0,0 +1,54 @@
...
+        right_msg.data = right_tread;
+        left_tread_publisher.publish(left_msg);
+        right_tread_publisher.publish(right_msg);
+    }
+
+    // left_tread and right_tread are output parameters; the rest are inputs.
+    void DrivebasePublisher::twistToDifferential(float linear_v, float angular_v,
+        float wheel_radius, float axle_length, float& left_tread, float& right_tread)
+    {
+        // Desired velocity across the ground for each wheel
```
AdamuKaapan: Not super necessary, especially if we're doing default values, but asserting `wheel_radius != 0` might be a good idea here since we're dividing by it (even though there should never be a case when it *is* zero).  While we're at it adding nonzero assertions to axle_length might be a good idea too, even though if it is zero it won't crash anything (just probably won't work as intended).

src/tfr_control/include/tfr_control/drivebase_publisher.h
```diff
@@ -0,0 +1,57 @@
...
+
+        DrivebasePublisher& operator=(const DrivebasePublisher&) = delete;
+        DrivebasePublisher& operator=(DrivebasePublisher&&) = delete;
+        
+
+        // Making this static allows for unit-testing without 
+        // creating a DrivebasePublisher object
+        // left_tread and right_tread are output parameters; the rest are inputs.
+        static void twistToDifferential(float linear_v, float angular_v, float wheel_radius, 
+                                        float axle_length, float& left_tread, float& right_tread);
```
AdamuKaapan: `wheel_radius` and `axle_length` could be marked as `const` (see comment further in the pull request), as well as `linear_v` and `angular_v`, just to indicate for sure that they're only inputs and not meant to be modified. Not required, just makes some things clearer.

src/tfr_control/test/test_tfr_control.cpp
```diff
@@ -4,6 +4,54 @@
...
+        DrivebasePublisher::twistToDifferential(2, 3.1415, WHEEL_RADIUS, 
+                              AXLE_LENGTH, left_tread, right_tread);
+        EXPECT_NEAR(3.040157823, left_tread, ABS_ERROR);
+        EXPECT_NEAR(24.34772304, right_tread, ABS_ERROR);
+
+        DrivebasePublisher::twistToDifferential(1.819346, 5.10293874, WHEEL_RADIUS, 
+                              AXLE_LENGTH, left_tread, right_tread);
+        EXPECT_NEAR(-4.848610462, left_tread, ABS_ERROR);
+        EXPECT_NEAR(29.76262621, right_tread, ABS_ERROR);
+
```
AdamuKaapan: Do we have tests for negative inputs? Super large/super large negative inputs? Tests for passing in NaN/Infinity?

AdamuKaapan (changes requested):
Added comments about test coverage, marking a few variables/parameters as `const`, some default value checks, and a few naming issues. Also raised the idea of a glossary/definition of terms somewhere.

collinsc (dismissed):

collinsc:
Fixed all the gripes I saw on first pass, adam does bring up some good ros stuff, I support what he mentioned.

src/tfr_control/include/tfr_control/drivebase_publisher.h
RyanBerge: We shouldn't need a definition of terms if our naming choices are descriptive enough.  I'm changing the name to `wheel_span` to try to address this.  Is that more intuitive?

src/tfr_control/include/tfr_control/drivebase_publisher.h
RyanBerge: Actually, function and method naming conventions was changed to `lowerCase`. I just keep forgetting to merge it.

src/tfr_control/src/drivebase.cpp
RyanBerge: I agree.  What I'm not sure about is how severe such an error should be.  If the values don't come in, or come in wrong, the robot won't move as expected but it won't ruin everything.  That said, it should be easy for us to always make sure our inputs are always loaded.

src/tfr_control/test/test_tfr_control.cpp
RyanBerge: I'm not sure there's value in having tests for NaN or Infinity, since I'm not sure what the expected behavior should be.  Negative inputs are definitely legal, so I'll add a couple for that.

src/tfr_control/include/tfr_control/drivebase_publisher.h
AdamuKaapan: Somewhat, I think `wheel_center_distance` might make a bit more sense. It's kinda a weird term to describe in a variable name

src/tfr_control/include/tfr_control/drivebase_publisher.h
AdamuKaapan: Ah, right, okay.

src/tfr_control/src/drivebase.cpp
AdamuKaapan: If we set the defaults to what we'd probably be feeding it anyways, I think warning or even info (depending on how much we care).

src/tfr_control/test/test_tfr_control.cpp
AdamuKaapan: My personal opinion would be having it stop all movement (so return 0) probably with a ROS_ERROR output, but there's room for debate on that.

RyanBerge:
@AdamuKaapan @collinsc I've finished adding in the changes for the second pass, most of which I agree with.

I have the parameter thing set to be an error right now because it falls more in line with the "Design by Contract" methodology (even if we aren't totally following that), but I'm open to hearing more opinions on that one.

AdamuKaapan:
Yeah, that's totally fine, makes sure if something goes wrong we'll see it. I'll get around to reviewing the rest later tonight.

collinsc (approved):
looks good!

AdamuKaapan (approved):

PR #12, opened 2018-01-18
Pc imaging/Hectorslam

collinsc:
# Description
The original intention of this commit was to get rtabmap setup. I did, and the stereo mapping utilites were of insufficient quality so I set up hector slam as a replacement. It's very responsive, and accurate at determining vertical obstacles, and relatively close ranges. It's slam functionality is terrible, but, provided with fused odometery it should be fine.

I'lll make a recommendation to the team that if accurate slam is a must the kinect v1/2 will be our be our most cost effective and accurate option.

I'll delve into detail what I tested, and how in further sections.

# Reviewer Evaluation Guide
My work relies on hardware so you will not be able to run it. The main entry point on

The bulk of my work was modifying anything with the prefix hector. I would recommend looking it over and making sure I am not doing anything insane in a launch file.

I also did a cleanup of erroneous files and dependencies in the navigation, launch,  and sensor package.
This included creating the core.launch file in tfr_launch, which extracted robot level XML actions to happen at launch file startup.

# Testing Notes
I tried several things that eventually led me to the conclusion that hector slam was the way to go.
This section is for future maintenance.

## Rtab setup in stereo B
Rtabmap provides the option to pipe rectified stereo images directly to the library, and it will take care of the processing. What I found was that the processing was really designed for cameras different than the duo, and had a wider seperation of lenses. This made it really not ideal, and it had a very restricted field of view and was not accurate

## RTab setup in stereo A
Rtabmap provides a configuration for setting up the library with a fudged depth image. Where you feed it a disparity image and it spits out an rgbd style depth image. This would be fine, but the driver does not output a standard ros stereo_msgs/DisparityImage, it outputs a ros sensor_msgs/Image, which is missing some important meta data and camera information.

## Disparity Image Wrappper
To solve the previously mentioned data incompatibility problem, I noticed that you can create a Disparity image from the data ros provided by writing a node that would perform the conversion. Really it just packaged the untouched disparity image with the calibration parameters of the duo. I wrote this in c++ and it is in one of the commits I am turning in. This worked, but on testing it turns out duo uses a different encoding scheme for it's images than rtabmap expects.

## Modifying the duo driver
I looked up some tutorials on image encoding and tried modifying the duo driver to match what rtabmap expected. This worked, but the resulting encoded image was not quite high enough quality for rtabmap to do any real processing. So this was a dead end.

## gmapping
I looked up alternative slam packages and gmapping came up right away. It was super buggy and not designed for real time use.

## Point cloud filtering
I experimented with the idea of trying working with the generated point cloud instead of the stereo images, and that worked pretty good. I used the pointcloud_to_laserscan package, and it does a good job of filtering noise and detecting vertical obstacles.

## Hector Slam
I finally set up hector slam, and it worked pretty well. So I tuned a few things, took a pause and decided to move on.


src/tfr_launch/launch/hector_test.launch
```diff
@@ -0,0 +1,6 @@
+<launch> 
+	<include file="$(find tfr_launch)/launch/core.launch"/>
```
Jebbs: Tabs?

src/tfr_launch/launch/sensor_test.launch
```diff
@@ -1,4 +1,5 @@
 <launch>
-    <include file="$(find tfr_sensor)/launch/sensor_platform.launch"/>
-    <include file="$(find tfr_visualization)/launch/sensor_platform.launch"/>
+	<include file="$(find tfr_launch)/launch/core.launch"/>
```
Jebbs: More tabs

Jebbs (changes requested):
You've got a coupe of tabs, but other than that LGTM.

src/tfr_launch/launch/sensor_test.launch
collinsc: done!

src/tfr_launch/launch/hector_test.launch
collinsc: done!

Jebbs (approved):
👍

PR #13, opened 2018-01-20
MoveIt! library setup

AdamuKaapan:
# Goals
 * Add a system to simulate the hardware interface in rviz so that we can visualize MoveIt! commands
 * Set up the MoveIt! library for basic motion planning

# Works Completed
 * `controller_launcher` now also loads the joint limits from the robot URDF file
 * `controller` now has an option to use fake values, which essentially applies the command directly to the velocity and position (like if the robot were in an ideal world)
 * Added the auto-generated MoveIt files to a new package, `tfr_moveit` (this is in a separate package because there's a very large quantity of files and I didn't want to bog down the control package)
 * The tfr_control launch file `test_moveit.launch` launches the hardware interface nodes, the MoveIt node "move_group" (which provides our code interface to the motion planning library), and opens an rviz window to send planning commands.

# Results
Motion of all actuatable joints can now be visualized in rviz upon sending data to the corresponding `command` topic (such as `/left_tread_velocity_controller/command`). The MoveIt library can now receive commands through rviz to go to randomly created positions and execute on those plans.

The next steps after this pull request is to create the Arm Action Server, which will consume the API provided by `move_group` and send commands.

# Testing
To test, simply run `test_moveit.launch`. In the rviz window that appears, go to the MotionPlanning display and click on the "Planning" tab. Then, under "Query->Select Goal State", press "Update" (the dropdown should be set to "<random valid>". You should see an orange version of the arm appear in a new location. After you see this, you can press "Plan" and see the plan that MoveIt creates and then run it by pressing "Execute", or you can press "Plan and Execute" and it'll do both steps at once.


bradford-halter (dismissed):
Everything looks good!

src/tfr_control/src/controller.cpp
```diff
@@ -35,6 +35,16 @@ namespace tfr_control
 
     void Controller::write() {
         // TODO: Waiting on further hardware development to implement
+        if (use_fake_values) {
+            // Update all actuators velocity with the command (effort in).
+            // Then update the position as derivative of the velocity over time.
+            for (int i = 0; i < 7; i++) {
```
aabdin02: What is this 7 times the loop does based on? What will happen if we change the number of sensors? Is this the only place we use the sensor numbers?

aabdin02 (changes requested):
Everything look OK except for the one comment!

src/tfr_control/src/controller.cpp
AdamuKaapan: @aabdin02 The 7 is based on the number of actuators we have registered with the controller_manager. If we were to change the number of sensors/actuators, this number would also have to change. As this number is set in stone by design, I'm not worried about that here. There may be an incoming issue with the implementation of the hardware itself later but that shouldn't affect this (it's only on the hardware side, not with registering with controller_manager), and that's reserved for the second hardware interface card.

aabdin02:
Wouldn't it be better if we had a variable to track that?

src/tfr_control/src/controller.cpp
aabdin02: Wouldn't it be better to use a variable instead? It is something simple that could easily be forgetten

AdamuKaapan:
I think that since it's hard-set in design, it can go either way, it makes basically no change, although I think that for clarity (since, hey, at least one person had a question about what it meant) a named constant somewhere would be a good addition. I'll go ahead and do that

aabdin02 (approved):

aabdin02 (dismissed):
Awesome, Awesome!!!

PR #14, opened 2018-01-24
Added the Aruco Action Server

LinuxKid21:
# Purpose
To integrate the openCV Aruco library as an action server in ROS

# What Got finished and what needs to be done still
The server is completely finished as is an example client. Right now there is no way to generate aruco boards and the position of the markers is done manually in the code. Aruco also has quite a few parameters, but I did not create a way to change those on the fly because it didn't seem necessary (in fact, the defaults seem to work the best anyway).
Some code is being repeated that only have to be done once, but I will fix that in a future poll request.

# Testing
You will need to print out the image file in the tfr_aruco folder (each marker will be 3.2 cm). I find it helpful to mark the top left corner when its done by putting an arrow to it.
You will need a webcam as well.
If you can't print or don't have a webcam then I will be happy to send you recorded messages(as soon as I figure out how to do that)
After building the program, run the following:
rosrun uvc_camera uvc_camera_node
rosrun tfr_aruco tfr_aruco_server
rosrun tfr_aruco tfr_aruco_example_client
I think these will run without camera calibration. If not then you may have to follow the uvc_camera calibration tutorial or just ask me for aforementioned recorded messages.


collinsc:
@LinuxKid21  I'll start my review. In the future if you need to spawn a series of nodes in sequence launch files are a little more maintainable and testable.

However this time it should be fine.

collinsc:
@AdamuKaapan do you have any idea why jenkins isn't running on this node?

src/tfr_msgs/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ add_compile_options(-std=c++11)
 find_package(catkin REQUIRED COMPONENTS
```
collinsc, line 6: @LinuxKid21 @AdamuKaapan uvc_camera is an additional dependency that needs to be listed.

src/tfr_aruco/package.xml
```diff
@@ -0,0 +1,68 @@
...
+  <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>actionlib</build_depend>
+  <build_depend>roscpp</build_depend>
+  <build_depend>tfr_msgs</build_depend>
+  <build_export_depend>actionlib</build_export_depend>
+  <build_export_depend>roscpp</build_export_depend>
+  <build_export_depend>tfr_msgs</build_export_depend>
+  <exec_depend>actionlib</exec_depend>
+  <exec_depend>roscpp</exec_depend>
+  <exec_depend>tfr_msgs</exec_depend>
```
collinsc, line 60: add uvc_camera

src/tfr_msgs/CMakeLists.txt
collinsc: It also needs to be added to the readme

collinsc:
uvc_camera is deprecated, in a future commit let's move to libuvc it's successor

collinsc:
image

Detects image, but no location information. Is this expected behavior?

Computation graph:


image

src/tfr_aruco/src/tfr_aruco_example_client.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include <tfr_msgs/ArucoAction.h> // Note: "Action" is appended
+#include <actionlib/client/simple_action_client.h>
+
+typedef actionlib::SimpleActionClient<tfr_msgs::ArucoAction> Client;
+
+// Called every time feedback is received for the goal
+void feedback(const tfr_msgs::ArucoFeedbackConstPtr& feedback)
+{
+}
```
collinsc, line 18: If feedback isn't required it's better not to include it

src/tfr_aruco/src/tfr_aruco_example_client.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+        cameraImage = *msg;
+
+        // Members of ArucoGoal are defined in tfr_msgs/action/Action.action
+        tfr_msgs::ArucoGoal goal;
+        goal.image = cameraImage;
+        goal.camera_info = cameraInfo;
+
+        // Callback functions: Result, Start, Feedback
+        client.sendGoal(goal, &finished, NULL, &feedback);
```
collinsc, line 88: yeah you can just pass in another  null param here for feedback

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <actionlib/server/simple_action_server.h>
+
+
+typedef actionlib::SimpleActionServer<tfr_msgs::ArucoAction> Server;
+
+// This is the method that will be called when a client makes use
+// of this server. The provided goal is the "input".
+void execute(const tfr_msgs::ArucoGoalConstPtr& goal, Server* server)
+{
+    cv::Ptr<cv::aruco::Dictionary> dictionary = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_5X5_250);
```
collinsc, line 20: You do not check for preemption at the beginning of an action, this needs to be fixed.

collinsc (changes requested):
So for sure there is a dependency issue, and this needs to be automated into a launch file for sure. Also we need to delete spurious callbacks/ check for preemption when appropriate. The way the action server is set up right now is not very maintainable, it needs to be refactored into a class at some point. I'll leave it up you if that is this pull request or the next one when you figure out what you want to do with the board.

src/tfr_msgs/CMakeLists.txt
LinuxKid21: @collinsc Well it sound like we should use libuvc instead. Also, I don't think it should be a dependency of my package. It takes in sensor_msgs/Image and sensor_msgs/CameraInfo which are produced by any camera package(or should be produced by any camera package).

LinuxKid21:
@collinsc As for it saying 0, I'm not sure. I wonder if the Aruco function is refusing to work because you don't have the camera configured. (The warning in the message) Can you please try configuring your camera or use this [yaml](https://drive.google.com/open?id=1Uuua1hNISJEKH4EU6HZp8zrt0nE6DvAX) file? (if you use mine, then it should just return incorrect results, but something not 0. Also it will give a separate error saying that it looks like a different camera than the yaml).

LinuxKid21:
@collinsc @AdamuKaapan Also, as far as Jenkins not running, I wonder if it's because I cloned the repository and tried merging that instead of creating a branch in this repository and merging that. I can close this commit and redo it that way if necessary.

RyanBerge:
@LinuxKid21 It probably would be best to make your changes in a branch off of this repository.  I'm not familiar enough with Git to know what the workflow looks like your way.

src/tfr_msgs/CMakeLists.txt
collinsc: Well when you write the launch file automating the code it will be.

collinsc: It will be a runtime dependency.

collinsc:
@LinuxKid21 The reason I think it is a bug, is that when it doesn't detect any markers and it's not near the sheet a string is displayed to the screen saying no markers found. When it get's close it displays that data, but all zeros. When I move it away it goes back to the no markers found bit.

I don't really care if it runs on my machine, but as long as it works on yours and the jetson robustly we are fine. I just can't look at your data.

Also how is it reporting data? Is this data saying the bin is this far away and this rotation from my reference frame? or is it saying I am this far away and at this rotation from the bin?

I think when we originally talked this was going to be a geometry_msgs/Pose? I think that would be a lot better. http://docs.ros.org/api/geometry_msgs/html/msg/Pose.html 

It will increase interoperability with the transform system and make math simpler.

collinsc:
@LinuxKid21 @RyanBerge  I also agree with ryan on this one, our tools and build server wont work with the fork approach.

LinuxKid21:
Okay. I'll close this later and do it the right way. In the mean time, still leave comments here.

collinsc:
@LinuxKid21 
Cool Also another thing that came to me, what was the purpose in seperating this from the rest of the localization code in it's own package?

LinuxKid21:
Hmmm... There's not a particular reason that they shouldn't be together. It just seemed better to separate them because the aruco code is able to it's own module. For example, putting it together would mean that all localization code would require the aruco module dependencies in its CMake. What does @Jebbs think?

src/tfr_aruco/package.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>tfr_aruco</name>
+  <version>0.0.0</version>
+  <description>The tfr_aruco package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="levi@todo.todo">levi</maintainer>
```
Jebbs, line 10: You probably want to put in your full name here.

src/tfr_aruco/package.xml
```diff
@@ -0,0 +1,68 @@
...
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
Jebbs, line 16: Don't we have a license?

src/tfr_aruco/package.xml
```diff
@@ -0,0 +1,68 @@
...
+  <!-- Url tags are optional, but multiple are allowed, one per tag -->
+  <!-- Optional attribute type can be: website, bugtracker, or repository -->
+  <!-- Example: -->
+  <!-- <url type="website">http://wiki.ros.org/tfr_aruco</url> -->
+
+
+  <!-- Author tags are optional, multiple are allowed, one per tag -->
+  <!-- Authors do not have to be maintainers, but could be -->
+  <!-- Example: -->
+  <!-- <author email="jane.doe@example.com">Jane Doe</author> -->
```
Jebbs, line 28: Everyone who worked on this should be listed here.

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,134 @@
...
+        marker.push_back(marker[0] + cv::Point3f(0,   3.2, 0));
+        
+        // from cm to meters
+        for(auto &p : marker) {
+            p *= .01;
+        }
+    }
+    std::vector<int> boardIds = {1, 2, 3, 4, 5, 6, 7, 8};
+    
+    cv::Ptr<cv::aruco::Board> board = cv::aruco::Board::create(std::move(boardCorners), dictionary, std::move(boardIds));
```
Jebbs, line 56: If the aruco board is always the same (or is always going to be the same), then it should be predefined and made constant.

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,134 @@
...
+    cv::Mat distCoeffs = cameraModel.distortionCoeffs().clone();
+    
+    std::vector< cv::Vec3d > rvecs, tvecs;
+    cv::aruco::estimatePoseSingleMarkers(markerCorners, 0.05, cameraMatrix, distCoeffs, rvecs, tvecs);
+
+
+
+    
+    cv::Vec3d boardRotVec, boardTransVec;
+    int v = cv::aruco::estimatePoseBoard(markerCorners, markerIds, board, cameraMatrix, distCoeffs, boardRotVec, boardTransVec);
```
Jebbs, line 105: `v` is not very descriptive as a variable name.

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,134 @@
...
+    
+    tfr_msgs::ArucoResult result;
+    result.number_found = v; // if v = 0, then no marker was detected
+    result.position = std::vector<float>(3, 0); // 3 members initialized to 0
+    result.rotation = std::vector<float>(3, 0); // 3 members initialized to 0
+    if(v != 0)
```
Jebbs, line 111: Will `v` ever be less than zero? If not, then `if (v>0)` might read better. This would be even more true with a different variable name.

I would also move the `// if v = 0, then no marker was detected` comment to be referencing this if statement.

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+    // using cameraImage is the same as cv_ptr->image, but more convienent and clear 
+    cv::Mat cameraImage = cv_ptr->image;
+
+
+    // detect fiducial markers
+    std::vector<int> markerIds;
+    std::vector<std::vector<cv::Point2f> > markerCorners;
+    
+    cv::aruco::detectMarkers(cameraImage, dictionary, markerCorners, markerIds, params);
```
Jebbs, line 91: You say that using `cameraImage` is the same as using `cv_ptr->image`, but you only use it once. You could get the same clarity by renaming `cv_ptr` to something more meaningful, like `cameraData`. Then you could use `cameraData->image` and it is just as clear.

Jebbs (changes requested):
Looks good for the most part save for my suggestions. The one about `cameraImage` vs `cv_ptr->image` is more of my own opinion than a change I require, though.

There's also a lot of newlines in here which I think could be trimmed up a little, but maybe that is also just my opinion.

I didn't run any tests myself since @collinsc did and found errors, but using a recording to do the testing is mentioned in the pull request description. Could we add that as a test that could be automated?

src/tfr_aruco/package.xml
collinsc: BSD baby! Let's put that in there

collinsc:
Hey I got a version of this branch working on the jetson w/ a launch file. I needed a way to test the rear camera I set up and it works pretty decent. So I don't know what was going on with my laptop, but it works with the jetson, which is what matters at the end of the day.

PR #15, opened 2018-01-24
Navigation Action Server

collinsc:
# Navigation Action Server
Contains the current client interface and "plumbing" of the navigation action server.

It consists of two nodes that wrap some encapsulating objects:

- Navigation Action Server:
  - The action  server that responds to user autonomous navigation requests. It contains a helper object called the navigation goal manager that manages goal state and goal updates. I did a separation of concerns here to make it more easily testable.
- Navigation Action Client:
  - This was a test node designed to show off the functionality of  it's contained Action Client and how it interacts with the action server. The idea being that later we can reuse the public interface of the action client in the executive node. What it tests and how will demonstrated in the testing section.

As an aside, all of the locations in the code that need further integration, or deletion when the navigation stack, or bin tf publisher is online, are given a `TODO` or `DELETE` comment annotation.

## Testing
There are ample unit tests for the navigation goal manager because it has the most mutable state, and does not rely on ros message passing.

There is a little bit of manual testing. I included a test launch file to demonstrate functionality.  Call it with `roslaunch tfr_launch navigation_as_test.launch`

Here is what it is supposed to do in sequence.
1. Spin up and action server
2. Spin up an action client
3. Start a goal
4. Preempt  that goal
5. Start a digging goal
6. Get feedback
7. Complete digging goal
9. Start Dumping goal
10. Get feedback
11. Finish Dumping goal

### DIFF
--- a/src/tfr_msgs/action/Navigation.action
+++ b/src/tfr_msgs/action/Navigation.action
@@ -1,12 +1,22 @@
 #goal msg
-uint8 cmd_code
-uint8 TO_MINING = 0 #Navigate to the mining zone
-uint8 TO_DUMPING = 1 #Navigate to the dumping zone
+Header header
+uint8 location_code
+uint8 UNSET = 0 #not specified yet by the user
+uint8 TO_MINING = 1 #Navigate to the mining zone
+uint8 TO_DUMPING = 2 #Navigate to the dumping zone
 ---
 #feedback msg
+Header header
 uint8 status
 uint8 OK = 0
+uint8 ERROR = 1
+geometry_msgs/Pose current
+geometry_msgs/Pose goal
 ---
 #result msg 
+Header header
 uint8 status
 uint8 OK = 0
+uint8 FAILURE = 1
+geometry_msgs/Pose current
+geometry_msgs/Pose goal
--- a/src/tfr_msgs/action/Navigation.action
+++ b/src/tfr_msgs/action/Navigation.action
@@ -1,9 +1,9 @@
 #goal msg
 Header header
 uint8 location_code
-uint8 UNSET = 0 #not specified yet by the user
-uint8 TO_MINING = 1 #Navigate to the mining zone
-uint8 TO_DUMPING = 2 #Navigate to the dumping zone
+uint8 UNSET = 0
+uint8 TO_MINING = 1
+uint8 TO_DUMPING = 2
 ---
 #feedback msg
 Header header
--- a/src/tfr_msgs/action/Navigation.action
+++ b/src/tfr_msgs/action/Navigation.action
@@ -1,0 +1,12 @@
+#goal msg
+uint8 cmd_code
+uint8 TO_MINING = 0 #Navigate to the mining zone
+uint8 TO_DUMPING = 1 #Navigate to the dumping zone
+---
+#feedback msg
+uint8 status
+uint8 OK = 0
+---
+#result msg 
+uint8 status
+uint8 OK = 0
--- a/src/tfr_navigation/CMakeLists.txt
+++ b/src/tfr_navigation/CMakeLists.txt
@@ -8,14 +8,10 @@ find_package(catkin REQUIRED COMPONENTS
   tfr_msgs
   geometry_msgs
   nav_msgs
-  std_msgs
   actionlib
 )
 
 find_package(GTest REQUIRED)
-
-catkin_package(
-)
 
 include_directories(
     include/${PROJECT_NAME}
@@ -23,28 +19,45 @@ include_directories(
     ${GTEST_INCLUDE_DIRS}
 )
 
-add_library(navigation_goal_manager src/navigation_goal_manager.cpp)
-target_link_libraries(navigation_goal_manager ${catkin_LIBRARIES}) 
+catkin_package(
+    INCLUDE_DIRS include/${PROJECT_NAME}
+    LIBRARIES ${PROJECT_NAME} navigation_server navigation_client
+    CATKIN_DEPENDS  roscpp tfr_msgs actionlib nav_msgs geometry_msgs
+)
+
+
+
+
+add_library( navigation_server
+    src/navigation_goal_manager.cpp
+    src/navigator.cpp
+)
+add_dependencies(navigation_server ${catkin_EXPORTED_TARGETS})
+
+add_library( navigation_client 
+    src/navigation_client.cpp
+)
+add_dependencies(navigation_client ${catkin_EXPORTED_TARGETS})
+
 
 add_executable(navigation_action_server 
     src/navigation_action_server.cpp
-    src/navigator.cpp
 )
-target_link_libraries(navigation_action_server ${catkin_LIBRARIES} navigation_goal_manager)
 add_dependencies(navigation_action_server ${catkin_EXPORTED_TARGETS})
+target_link_libraries(navigation_action_server navigation_server ${catkin_LIBRARIES})
+
 
 add_executable(navigation_action_client 
     src/navigation_action_client.cpp
-    src/navigation_client.cpp
 )
-target_link_libraries(navigation_action_client ${catkin_LIBRARIES})
 add_dependencies(navigation_action_client ${catkin_EXPORTED_TARGETS})
+target_link_libraries(navigation_action_client navigation_client ${catkin_LIBRARIES})
 
 SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
 
 if (CATKIN_ENABLE_TESTING)
     catkin_add_gtest(${PROJECT_NAME}-test test/test_goal_setting.cpp)
-    target_link_libraries(${PROJECT_NAME}-test ${catkin_LIBRARIES} navigation_goal_manager)
+    target_link_libraries(${PROJECT_NAME}-test navigation_server ${catkin_LIBRARIES})
 endif()
 
 
--- a/src/tfr_navigation/CMakeLists.txt
+++ b/src/tfr_navigation/CMakeLists.txt
@@ -1,39 +1,33 @@
-# If you're having trouble getting your things to build, visit the ROS
-# tutorials at <http://wiki.ros.org/ROS/Tutorials> and also read the ROS
-# CMake documentation at <http://wiki.ros.org/catkin/CMakeLists.txt>.
-
 cmake_minimum_required(VERSION 2.8.3)
 project(tfr_navigation)
 
 add_compile_options(-std=c++11)
 
-find_package(
-    catkin REQUIRED COMPONENTS
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+  tfr_msgs
+  actionlib
 )
 
 find_package(GTest REQUIRED)
 
-# These are all for exporting to dependent packages/projects.
-# Uncomment each if the dependent project requires it
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES tfr_sensor
-#  CATKIN_DEPENDS roscpp std_msgs
-#  DEPENDS system_lib
 )
 
-# Specify additional locations of header files
-# Your package locations should be listed before other locations
 include_directories(
-  include/${PROJECT_NAME}
   ${catkin_INCLUDE_DIRS}
   ${GTEST_INCLUDE_DIRS}
 )
 
-# This call is sometimes needed and sometimes not and I'm not really clear why
+
+
+
+add_executable(navigation_action_server src/navigation_action_server.cpp)
+target_link_libraries(navigation_action_server ${catkin_LIBRARIES})
+add_dependencies(navigation_action_server ${catkin_EXPORTED_TARGETS})
+
 SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
 
-# Add gtest based cpp test target and link libraries
 if(TARGET ${PROJECT_NAME}-test)
   target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME}_library)
 endif()
--- a/src/tfr_navigation/CMakeLists.txt
+++ b/src/tfr_navigation/CMakeLists.txt
@@ -20,12 +20,7 @@ include_directories(
 )
 
 catkin_package(
-    INCLUDE_DIRS include/${PROJECT_NAME}
-    LIBRARIES ${PROJECT_NAME} navigation_server navigation_client
-    CATKIN_DEPENDS  roscpp tfr_msgs actionlib nav_msgs geometry_msgs
 )
-
-
 
 
 add_library( navigation_server
@@ -45,7 +40,6 @@ add_executable(navigation_action_server
 )
 add_dependencies(navigation_action_server ${catkin_EXPORTED_TARGETS})
 target_link_libraries(navigation_action_server navigation_server ${catkin_LIBRARIES})
-
 
 add_executable(navigation_action_client 
     src/navigation_action_client.cpp
--- a/src/tfr_navigation/CMakeLists.txt
+++ b/src/tfr_navigation/CMakeLists.txt
@@ -6,6 +6,9 @@ add_compile_options(-std=c++11)
 find_package(catkin REQUIRED COMPONENTS
   roscpp
   tfr_msgs
+  geometry_msgs
+  nav_msgs
+  std_msgs
   actionlib
 )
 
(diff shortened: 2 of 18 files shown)


src/tfr_msgs/action/Navigation.action
```diff
@@ -0,0 +1,22 @@
...
+geometry_msgs/Pose current
+geometry_msgs/Pose goal
+---
+#result msg 
+Header header
+uint8 status
+uint8 OK = 0
+uint8 FAILURE = 1
+geometry_msgs/Pose current
+geometry_msgs/Pose goal
```
RyanBerge, line 22: I don't understand the format here:
```
uint8 <status_code>
uint8 CODE1 = 0
uint8 CODE2 = 1
uint8 CODE3 = 2
```
It seems like we're sending a byte for status code, which sounds fine, but then we're also sending the values for what that status code means?  I assume the `=` operator is for default values.  Is there a reason to send this extra information?

src/tfr_navigation/CMakeLists.txt
```diff
@@ -1,38 +1,64 @@
...
+
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES tfr_sensor
-#  CATKIN_DEPENDS roscpp std_msgs
-#  DEPENDS system_lib
+    INCLUDE_DIRS include/${PROJECT_NAME}
+    LIBRARIES ${PROJECT_NAME} navigation_server navigation_client
+    CATKIN_DEPENDS  roscpp tfr_msgs actionlib nav_msgs geometry_msgs
 )
```
RyanBerge: Does `tfr_navigation` show up as a dependency for any other package?  If not, these lines in `catkin_package` are not required; this statement specifies what the package exports for other dependent packages.

src/tfr_navigation/src/navigation_client.cpp
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * action - The byte code corresponding to a location in the world map
+ * */
+void NavigationClient::navigate(uint8_t location)
+{
+    tfr_msgs::NavigationGoal goal;
+    goal.location_code= location;
+    client.sendGoal(goal,
+            boost::bind(&NavigationClient::finished,this, _1,_2),
+            Client::SimpleActiveCallback(),
```
RyanBerge: What does the `SimpleActiveCallback()` function do?  I see it's a member of the `SimpleActionClient` class provided by ROS.  Do we need a function for start callbacks?

src/tfr_navigation/include/tfr_navigation/navigation_goal_manager.h
```diff
@@ -0,0 +1,34 @@
...
+class NavigationGoalManager
+{
+    public:
+        //dist_landing + dist_obstacle + 1/2 robot_length
+        constexpr static double SAFE_MINING_DISTANCE { 1.5 + 2.98 + 1.3/2 };
+        //mining_zone_width - digging_radius
+        constexpr static double MINING_LINE_LENGTH { 3.78 - 1.75 };
+        //length of landing zone
+        constexpr static double FINISH_LINE { 1.5 };
+
```
RyanBerge: I would like to see these values set up as inputs to the node so that we don't have to recompile later if we decide we want to tune them.

src/tfr_navigation/include/tfr_navigation/navigation_goal_manager.h
```diff
@@ -0,0 +1,34 @@
...
+        NavigationGoalManager(const NavigationGoalManager&) = delete;
+        NavigationGoalManager& operator=(const NavigationGoalManager&) = delete;
+        NavigationGoalManager(NavigationGoalManager&&) = delete;
+        NavigationGoalManager& operator=(NavigationGoalManager&&) = delete;
+
+        void initialize_goal();
+        void update_mining_goal(geometry_msgs::Pose msg);
+
+        //delegate initialization to ctor
+        uint8_t location_code;
```
RyanBerge: I saw this code in the client as well; is a single byte really sufficient to represent any coordinate we decide to move to?  It's only 256 positions----16x16 for a 2D space.  Unless I'm misunderstanding what this is for?

RyanBerge: Ah, I see now.  `TO_MINING` or `TO_DUMPING`.

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * */
+void NavigationGoalManager::initialize_goal() {
+    nav_goal.target_pose.header.stamp = ros::Time::now();
+    //TODO integrate the bin reference frame here, upon completion of
+    //the bin transform publisher
+    switch(location_code)
+    {
+        case(tfr_msgs::NavigationGoal::TO_MINING):
+            nav_goal.target_pose.pose.position.x = SAFE_MINING_DISTANCE;
+            //TODO
```
RyanBerge: I assume this is the place where we write in the logic for mapping the intermittent path?

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -0,0 +1,61 @@
...
+        double v_position = msg.position.y; 
+        int sign = (v_position > 0) ? 1 : -1;
+        nav_goal.target_pose.pose.position.y = sign*std::min(
+                MINING_LINE_LENGTH/2, std::abs(v_position));
+    }
+    else 
+    {
+        ROS_WARN("only can update mining goal, your goal: %d", location_code);
+    }
+}
```
RyanBerge: You have a misaligned brace in the `if` statement.

Is the `Pose` input the current robot state?  What is the reference coordinate system being used here?  I remember seeing a photo on Slack but I can't find it now.

src/tfr_navigation/include/tfr_navigation/navigator.h
```diff
@@ -0,0 +1,58 @@
...
+#ifndef NAVIGATOR_H
+#define NAVIGATOR_H
+#include<ros/ros.h>
+#include<ros/console.h>
+#include<actionlib/server/simple_action_server.h>
+#include<navigation_goal_manager.h>
+#include<tfr_msgs/NavigationAction.h>
+#include<nav_msgs/Odometry.h>
+#include<boost/bind.hpp>
+#include<cstdint>
```
RyanBerge: There are some missing spaces here. Doesn't cause any errors, but it does look weird.

src/tfr_navigation/src/navigator.cpp
```diff
@@ -0,0 +1,135 @@
...
+Navigator::Navigator(ros::NodeHandle &n,std::string name) : node{n},  server{n, name,
+    boost::bind(&Navigator::navigate, this, _1) ,false}, action_name{name},
+    goal_manager()
+{
+    ROS_DEBUG("Navigation server constructed %f", ros::Time::now().toSec());
+    //get parameters
+    ros::param::param<std::string>("~odometry_topic", odometry_topic,
+            "/fused_odom");
+    ros::param::param<float>("~rate", rate, 1);
+    ros::param::param<std::string>("~frame_id", frame_id, "base_footprint");
```
RyanBerge, line 17: Is there a reason these parameters all begin with a `~` character?

src/tfr_navigation/src/navigator.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *      Navigation.action in the tfr_msgs package
+ *      -Pose describing our final position
+ *  NOTE NOT THREAD SAFE
+ * */
+void Navigator::navigate(const tfr_msgs::NavigationGoalConstPtr &goal)
+{
+    ROS_INFO("Navigation server started");
+
+    //start with initial goal
+    goal_manager.location_code = (*goal).location_code;
```
RyanBerge: You could also use `goal->location_code` here, which might be easier to read.

src/tfr_navigation/src/navigator.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Callback for updating the most recent position
+ * */
+void Navigator::update_position(const nav_msgs::OdometryConstPtr &msg)
+{
+    //get the pose without the covariance, not needed
+    //need to use shared pointer here for thread safety see:
+    //https://answers.ros.org/question/53234/processing-an-image-outside-the-callback-function/
+    current_position = *msg;
```
RyanBerge, line 133: This comment is a `TODO:`, right?  Why not convert it to a thread-safe type now so we don't forget?

RyanBerge (changes requested):
This looks like a pretty good start.  I've made a number of comments; most of them are questions for further clarification.  In general I'd like to see a bit more in terms of documentation for classes and functions so people can figure out what they're doing very quickly if they don't want to parse through the code.

The big thing is that I need to see the coordinate system being used before I can approve.  I remember there was a screenshot or a photo on Slack, but I can't seem to find it now, and we should get that into a more formal part of our documentation so it can be easily referenced.

src/tfr_msgs/action/Navigation.action
collinsc: Thats how you declare a message level constant those aren't acutally sent over the net. It gets saved in an enum like device in the namespace of the generated message header.

src/tfr_navigation/CMakeLists.txt
collinsc: I had some issues with that line with that library that was marked as a library, I'll try deleting it, but I was not able to get my other code to build that reference external executables within the package without that list of dependencies

src/tfr_navigation/src/navigation_client.cpp
collinsc: It's a dummy function supplied by the library implementers its the equivalent of null.

src/tfr_navigation/include/tfr_navigation/navigation_goal_manager.h
collinsc: personally I think it's overkill but sure.

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: Pose is a relative measure of distance based upon our coordinate frame in the tf system

src/tfr_navigation/src/navigator.cpp
collinsc: yes because they are in our private namespace.

src/tfr_navigation/src/navigator.cpp
collinsc: oh sure I was silly

src/tfr_navigation/src/navigator.cpp
collinsc: No I am dereferencing a boost shared pointer. This comment is to let you know that this is a fully thread safe operation, and no maitenence is required

collinsc:
@RyanBerge It use stamped poses the coordinate system is built into the data. No further work needed.

src/tfr_msgs/action/Navigation.action
RyanBerge: Ah.  That's fine, then.

src/tfr_navigation/CMakeLists.txt
RyanBerge: Builds fine for me.  Give it another try?

I don't think having these lines in there will break anything, but it will slow down the compilation process a little.

src/tfr_navigation/src/navigation_client.cpp
RyanBerge: Might it be better to use `NULL`, then? It's a bit more clear as to what's happening.

src/tfr_navigation/CMakeLists.txt
collinsc: Fixed it! so in the future we fill this in only with the stuff that is going to be  used by external packages. I'll stick to that.

src/tfr_navigation/src/navigation_client.cpp
collinsc: Well I got this example from the tutorials on ros. I looked at the source code and it looks like this actually does something for that method a little more complicated than that, so I don't know if changing that type will be free of consequences.

Source: http://docs.ros.org/jade/api/actionlib/html/simple__action__client_8h_source.html

So how about I document with a comment explaining whats going on and leave it as is?

src/tfr_navigation/include/tfr_navigation/navigation_goal_manager.h
collinsc: Done! Retained constness as well, but don't get any of the benefits of constexpr and compile time checking anymore.

src/tfr_navigation/include/tfr_navigation/navigator.h
collinsc: done!

src/tfr_navigation/src/navigator.cpp
collinsc: Done!

collinsc:
@RyanBerge your changes have been incorperated

src/tfr_navigation/src/navigation_goal_manager.cpp
RyanBerge: This file still has a couple misaligned braces.

More importantly, I still don't understand the `get_updated_mining_goal` function.  I need to know our coordinate system so I can verify that the function is correct.

RyanBerge:
@collinsc  Changes look good, but the reason I need the coordinate system is so I can verify that the update algorithm is correct.  Might be a good example of the kind of place that warrants extra comments.

RyanBerge (dismissed):

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Update the mining goal to the most efficient place on the mining.
+ * fails gracefully and warns the user if there is an error.
+ */
+move_base_msgs::MoveBaseGoal NavigationGoalManager::get_updated_mining_goal(geometry_msgs::Pose msg)
+{
+    if (location_code == tfr_msgs::NavigationGoal::TO_MINING){
+        nav_goal.target_pose.header.stamp =ros::Time::now();
+        //TODO integrate reference frame when bin tf publisher is finished
+        double v_position = msg.position.y;
```
Jebbs: Is `v_position` vertical position? If so, I'd rather vertical be spelled out or use something like `y_position` as it is more recognizable.

src/tfr_navigation/include/tfr_navigation/navigation_client.h
```diff
@@ -0,0 +1,48 @@
...
+#include <ros/ros.h>
+#include <ros/console.h>
+#include <actionlib/client/simple_action_client.h>
+#include <boost/bind.hpp>
+#include <cstdint>
+#include <tfr_msgs/NavigationAction.h>
+class NavigationClient
+{
+    public:
+        enum class Location : uint8_t
```
Jebbs, line 22: Any reason for this enum class? You always use `tfr_msgs::NavigationGoal::TO_MINING` or `tfr_msgs::NavigationGoal::TO_DUMPING` explicitly.

src/tfr_navigation/src/navigation_client.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Nonblocking call to Navigate to an arbirtrary defined place as defined in the Navigation
+ * Goal constants header.
+ *
+ * action - The byte code corresponding to a location in the world map
+ * */
+void NavigationClient::navigate(uint8_t location)
+{
+    tfr_msgs::NavigationGoal goal;
```
Jebbs, line 20: Is a `tfr_msgs::NavigationGoal` a `uint8_t`? If so, why not just have the parameter be named "goal" instead?

Jebbs (changes requested):
I think something needs to be done with the `Location` enum class(in `navigation_client.h`), whether it get's removed or get's used. Personally, I would rather use something like this enum class instead of the generated tfr_msgs::NavigationGoal thing.

Otherwise, I don't think I saw anything I wouldn't be OK with, though I made a couple of suggested changes.

src/tfr_navigation/include/tfr_navigation/navigation_client.h
collinsc: type safety. 
To be honest I kind of want to scrap the message level constants and just use the c++ enum. 
Can we do that on the next pull request?

src/tfr_navigation/src/navigation_client.cpp
collinsc: See above comment I want to replace this with a strongly typed c++ enum. on the next pull request.

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: I'll fix this

Jebbs (dismissed):

PR #16, opened 2018-01-26
Transforms for nav stack

collinsc:
# Bin Transformations
So yesterday I talked with adam, and noticed there was a missing finishing step to localization that we hadn't discussed previously. We get back a pose from the aruco action server describing where the bin is in relation to the rover, but that isn't integrated into tf2 where we can use it for the rest of the program.

So this pull request wrote all of the neccessary components to make our current action servers play nicely with tf, and form the missing piece of integration between localization, navigation, and tf.

I didn't know where to stick this, and I could see us using this concept later to keep track of the hole, so I stuck it in the utility package.

### Point Broadcaster
A generic point that broadcasts a static distance from a frame to a parent frame, see the header for a detailed description. Exposes a service as an interface that let's the user move the point around the fixed frame at runtime. I kept it general because it made more sense from a design standpoint, and we might want to keep track of other important points later.

I wrote a launch file that spins up one of these to keep track of the bin. 

### Tf Manipulator
This was a helper class to help out with the math of transforming the output from aruco, to the point of reference of a frame.  In our case the map. I originally just wanted a few methods to take care of this in a namespace, however because of some constraints in the tf2 library, it became a class. Read the code for the details

### Bin  Localizer
This was a test class that creates an imaginary point, and uses the Tf Manipulator to initialize the Point Broadcaster, and displays the arrangement to rviz.

##Testing

This was all plumbing, it didn't feel appropriate to unit test, also testing manually was really easy in rviz.

To test, run `roslaunch tfr_launch bin_broadcaster_test.launch` and make sure to enable viewing the transforms in rviz.

You will see the following:
image


RyanBerge (changes requested):
Because these changes are not meant to be final implementations for the TF transformations, I'm not inclined to nitpick over the details.  The one thing I am going to request is some more descriptive comment headers for the classes so that people looking over the code have a better idea of what it does.

src/tfr_utilities/include/tfr_utilities/placeholder_utility.h
```diff
@@ -1,9 +1,10 @@
...
 
 namespace tfr_utilities
 {
-    int Foo();
+        int Foo();
 }
 
-#endif
\ No newline at end of file
+#endif
```
AdamuKaapan, line 10: Is this file necessary for now? I'd personally love to save this until we know we need it (although that point is up for debate).

AdamuKaapan:
Only complaint I have is that I think the scale is a bit off (which doesn't matter) compared to the robot model (I'd basically just multiply the distance by 5 or 10). Aside from that, I personally think the headers are actually pretty good (@RyanBerge ), maybe just copy them into the .cpp files as well?

AdamuKaapan (dismissed):
Approved with a couple comments: one is nitpicky and not particularly relevant because it has to do with a temporary class, and the other is about file headers, which seem pretty good, but could also be copied into the .cpp files. It all looks good enough to be merged though.

src/tfr_utilities/include/tfr_utilities/placeholder_utility.h
RyanBerge: I would say no, probably; we can delete this if we want.

collinsc: Oh this is actually a pretty huge change it causes all the remaining placeholder code not compile anymore, let's make that a seperate pull request

collinsc:
@AdamuKaapan  @RyanBerge  I copied over the headers and put in a few more comments. Let me know if that's enough.

RyanBerge (approved):

AdamuKaapan (approved):

PR #18, opened 2018-01-27
Rtab redux

collinsc:
# Rtab Parameter Tuning
So after trying every trick in my linux cookbook I could not get the libraries to compile from source, the jetson is just too weird right now, and I don't think it's worth any more of my time.

So I delved deep, and optimized as much as I can with the downloaded binaries. Which is signifigant because we can spin at 3 hz comfortably and in a stable prolonged manner, which is more than we could say before.

My odometry is not good, so I cannot tune it any further until sensor fusion makes some progress. It just shakes too much in test to get reliable data, but obstacle detection is pretty good at about 0.5-1m which is probably good enough once we have stable odometry.

There isn't much to review on this commit, the final product is just the params and automation scripts.

It has undergone some history revisions


RyanBerge (dismissed):
I'm not inclined to try to review the DUO driver since I don't understand any of it anyway and most of that code isn't ours in the first place.  I'm also not inclined to go over the Python odometer emulator with a fine comb since it won't be there when the competition happens.  What I did look at looks good to me; I'll take your word for it that the testing you've done in the lab was sufficient for now.

src/tfr_visualization/config/mapping.rviz
```diff
@@ -66,11 +65,11 @@ Visualization Manager:
...
-          Value: true
+          value: true
         duo3d/camera_frame:
-          Value: true
+          value: true
         duo3d/frame:
-          Value: true
+          value: true
         duo3d/imu_frame:
           Value: true
```
Jebbs, line 74: Should the other `Value` variables be changed to lowercase `value`?

Jebbs (dismissed):
Everything I looked at looks fine. I have one comment about some variable names in a launch file, but only because some were changed, but not others.

src/tfr_visualization/config/mapping.rviz
collinsc: This is generated code, so I can't really change it, however in our system for our yaml, yes you are right.

collinsc:
fixed faulty merge, but no changes, merging with @Jebbs approval

PR #19, opened 2018-01-29
Create Mission Control Package

Jebbs:
### About:
This pull creates the Mission Control package, which supersedes and replaces the Communication package. The current implementation is simpley a ROS Node that listens for messages on a topic dedicated to status messages from the various subsystems: tfr_system_status.

Additionally, this adds the structure of the message the subsystems should use when communicating to Mission Control.

### Testing:
With the Mission Control node running, one can use `rostopic` to publish a message to the tfr_system_status topic. The node will see them and display the contents of the message to the console.

### Notes:
Some things in the CMake file were commented out instead of being removed because it is likely they will be needed in the future.

I also wasn't sure of the naming convention for topics, so please let me know if I should change it.


src/tfr_mission_control/src/mission_control.cpp
```diff
@@ -0,0 +1,30 @@
...
+  ros::init(argc, argv, "mission_control");
+
+  /**
+   * NodeHandle is the main access point to communications with the ROS system.
+   * The first NodeHandle constructed will fully initialize this node, and the last
+   * NodeHandle destructed will close down the node.
+   */
+  ros::NodeHandle n;
+
+  ros::Subscriber sub = n.subscribe("tfr_system_status", 1000, testCallback);
```
collinsc: That's a really big communication buffer, 10-50 might be a little more realistic.

collinsc: Aslo we never really established a uniform way of naming topics, but we haven't been using the tfr prefix. I think the spec initially named it something short like "com" or "status"

collinsc: What do you think? @Jebbs @RyanBerge

src/tfr_msgs/msg/SystemStatus.msg
```diff
@@ -0,0 +1,3 @@
+time SendTime
```
collinsc: So instead of sending a timestamp like this I am going to suggest we use the standard ros header object for consistency. Pretty much every top level msg data type in ros has one. 

A concrete benefit this gives us is that ros will automagically assign each message sent like this a unique seq number, and there is greater interoperability with existing messsage apis, if that ever becomes a priority.

It's described pretty well in this article in the header section. It's just a convention thing.
http://wiki.ros.org/msg

collinsc (changes requested):
I think that topic name should be shortened a little to fit with our naming convention for the rest of the system. Generally top level stuff is kept short, while more complicated stuff gets it's own namespace and names are still kept short.

Also I asked for a small additive change on the custom message type for greater ROS interoperability.

All in all good! let's just fix our foundation before building the house.

src/tfr_msgs/msg/SystemStatus.msg
collinsc: @Jebbs Disregard this! I just realized this is only applicable for things we want to integrate with tf, which we would never have reason to with this. so if you just want a timestamp for internal stuff, this is more appropriate.

I did some more reading and changed my mind.

src/tfr_mission_control/src/mission_control.cpp
RyanBerge: Some of our topics are generated by ROS components and those names are a little out of our control.  I don't think keeping it super short is important; I'd be happy with `system_status`.

collinsc (approved):

RyanBerge (approved):

PR #20, opened 2018-01-31
Light Detection

collinsc:
# Description

I wrote a node that detects sharp changes in light. The detection action server. 

It is started by giving it a goal, where it spins up and starts looking at a configurable camera feed. It keeps a moving average of color intensities (0->255) for each color channel. Currently when it detects a significant image intensity change it marks it's goal to complete as a signal to the action client and stops image processing. 

Some configurable paramters I left exposed to allow us to tweak sensitivity are:
  - average width: How long of a moving average
  - intensity threshold: How sharp of a spike is considered turning on the light.

Now because we don't have the led for this set up yet, and don't know what color it is. I don't have this set to detect any particular color, instead it just weighs them easily. This is a one line of code fix when we have the device so I am not particularly worried.

However the results are pretty good, at least on my webcam. So in theory they will get even more robust with the actual light.

# Testing
I wrote a test class `light_detection_action_client` to test this code, and show off it's functionality. I made use of the `cv_camera` to generate the images. In my case it found my laptop webcam out of the box. Let me know if you have any trouble getting an image, and I can help you troubleshoot.

The launch file can be run with `roslaunch tfr_launch light_test.launch`.

After running you should see the action server:
1. start
2. connect with the client
3. accept a goal
4. get preempted 
5. start a new goal
6. wait for a light

At this point you need to mimic a light being turned on, either turn on your phone flashlight next to the camera aperture, or cover and uncover it with your palm. 

7. a success message printed

There isn't much math we want to test here that's any more complicated than an average, all of the heavy lifting is done by the opencv matrix libraries.
If you want to see calculated intensities, uncomment line 74 of light_detector.cpp.

__EDIT:__
So there was an extra commit that was on here that sets it up on the jetson camera I added an extra launch file to test at home. Do:  `roslaunch tfr_launch local_light_test.launch`.


src/tfr_sensor/src/light_detection_action_server.cpp
```diff
@@ -0,0 +1,90 @@
...
+                return;
+            detector.add_image(msg);
+            if (detector.is_on())
+                server.setSucceeded();
+        }
+
+        ros::NodeHandle &n;
+        actionlib::SimpleActionServer<tfr_msgs::LightDetectAction> server;
+        LightDetector detector;
+        image_transport::ImageTransport it;
```
RyanBerge, line 67: I'd like a longer name for this variable.

src/tfr_sensor/src/light_detector.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ColorStats stats{};
+    auto scale = [&image] (const double &intensity) 
+    {
+        return intensity/(image.rows*image.cols);
+    };
+
+    //note we have to reverse out of native cv bgr ordering
+    stats.r_ave=scale(intensities[2]);
+    stats.g_ave=scale(intensities[1]);
+    stats.b_ave=scale(intensities[0]);
```
RyanBerge: In the future, let's make sure lambda expressions really make sense before we write them.  Most people, myself included, don't use them very often and are fairly unfamiliar with their syntax, which means it takes a long time to figure out what's going on.  Especially here, since it could have been as simple as `stats.r_ave = intensities[2] / (image.rows * image.cols);`.

src/tfr_sensor/src/light_detector.cpp
```diff
@@ -0,0 +1,85 @@
...
+     *Get an average of the color channels for every element exept for the most
+     recent addition to the average.
+     * */
+    std::for_each(++brightness.begin(), brightness.end(), 
+            [&total, &ave_size] (const ColorStats &stat)
+            {
+                total.r_ave += stat.r_ave/ave_size;
+                total.g_ave += stat.g_ave/ave_size;
+                total.b_ave += stat.b_ave/ave_size;
+            });
```
RyanBerge: See above comment.  Normal `for` loop or even a `ranged for` loop would be much better for readability here.

src/tfr_sensor/src/light_detector.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    //ROS_INFO("int: %f %f %f", total.r_ave, total.g_ave, total.b_ave);
+    ColorStats recent = brightness.front();
+
+    auto amalgamate = [](const ColorStats &c)
+    {
+        return (c.r_ave + c.g_ave + c.b_ave)/3;
+    };
+    double recent_ave = amalgamate(recent);
+    double total_ave = amalgamate(total);
```
RyanBerge: I'm detecting a pattern...

src/tfr_sensor/src/light_detector.cpp
collinsc: I would argue the lambda makes this more tedious calculation more readable. The name of the lambda tells you exactly what I am doing, scaling the image intensity. 

If I ever want to modify how I am scaling the image in the future, I just change the scaling function. Then it does all 3 channels at once. There is also like 0 extra performance overhead. 

Most of the modern 11, 14, and 17 standards make heavy use of them as a full blown replacements for functors, so lambdas are here to stay, and I think will become more and more prevalent as a useful construct.

src/tfr_sensor/src/light_detection_action_server.cpp
collinsc: I can do that.

src/tfr_sensor/src/light_detector.cpp
collinsc: This is just basic functional programming, here, and uses the standard libraries effectively. This is a perfectly valid use case for a foreach.

src/tfr_sensor/src/light_detector.cpp
collinsc: This is also a perfectly valid use case. It give an english description to math.

src/tfr_sensor/src/light_detector.cpp
RyanBerge: Replacements for functors, yes.  That is their purpose and they're a great way for generating functors.  You don't need functors here.

src/tfr_sensor/src/light_detector.cpp
RyanBerge: This is a terrible use of `for_each` because it can be replaced with a ranged-for loop without losing anything.

RyanBerge (changes requested):
You've heard my arguments on the lambdas.  I think you have a case (a weak one) for the first and third uses, but not for the `std::for_each` loop.  I'd like to hear some opinions from the rest of the team on the use of lambdas and where they might be appropriate.

Everything else looks fine.

src/tfr_sensor/src/light_detector.cpp
collinsc: well you couldn't use a ranged for, you would need to use an old fashioned for loop (with a little auto to cut down on iterator verbosity), and increment the start iterator like I do in the foreach. The issue is that we need to sum all of the elements except for the 1st element. The ranged for wouldn't expose the starting and ending iterators for me to offset.

so the alternative would be:
```
ColorStats total{};
int ave_size = brightness.size() -  1;

for(auto elem = ++brightness.begin(), auto end brightness.end(); elem != end; ++elem)
{
                total.r_ave +=  elem->r_ave/ave_size;
                total.g_ave += elem->g_ave/ave_size;
                total.b_ave += elem->b_ave/ave_size;
}
```

RyanBerge: You're right, you can't use ranged-for if you want to skip the first element.  I hadn't considered that.  But you don't have to use iterators here; you can just index into the data structure to get each element.

```
for (int i = 1; i < brightness.size(); i++)
{
    total.r_ave += brightness[i].r_ave / ave_size;
    total.g_ave += brightness[i].g_ave / ave_size;
    total.b_ave += brightness[i].b_ave / ave_size;
}
```

src/tfr_sensor/src/light_detection_action_client.cpp
```diff
@@ -0,0 +1,32 @@
...
+    ROS_INFO("client connected");
+    tfr_msgs::LightDetectGoal goal{};
+    
+    client.sendGoal(goal);
+    ros::Duration(0.5).sleep();
+    client.cancelAllGoals();
+    ros::Duration(0.5).sleep();
+
+    client.sendGoal(goal);
+    while (!client.waitForResult(ros::Duration(0.25)));
```
AdamuKaapan, line 27: Do we have anything to handle if we don't see the light for a very long time? While obviously this would be very bad for this to happen in a match (it means either our light hardware broke or we're searching in a veeery wrong place), it might be good to have a timeout (even if when it times out we just run it again and output a debug message) just in case so that we don't get stuck waiting forever? Just a thing I noticed and figured I might bring up.

AdamuKaapan: I'm thinking just adding a boolean to the action result saying whether we saw it or not?

src/tfr_sensor/src/light_detector.cpp
AdamuKaapan: (looked through all the cases where this happened before writing this response): I feel like this case could go either way in my opinion, although I do ask that if we stay with a lambda I think we might want a different name than "scale". For me "scale" implies resizing the image based on the work I've done in imaging programs, and given that we're working with images here that makes it more confusing (for me at least).

src/tfr_sensor/src/light_detector.cpp
AdamuKaapan: I think I agree with @RyanBerge for this case, a for loop like he wrote is a lot more readable to me than the for_each call.

src/tfr_sensor/src/light_detector.cpp
AdamuKaapan: Again, I feel like this could go either way like the first case, although the name is definitely much better in this case. Whatever we decide for either one though we should stick with consistency, because this is almost the exact same as the first case.

AdamuKaapan (changes requested):
Added comments onto the lambda expression dispute, and also added a concern/question about having a possibly infinitely running (?) light searching method.

AdamuKaapan:
I guess my summary on the lambda dispute is this: I agree with @RyanBerge that the for_each call is pretty unreadable and can be easily replaced with a standard for loop. The first and third instances I'm relatively neutral on: the lambda syntax isn't too hard to comprehend (although somewhat worse than standard functions) and *given a proper name is chosen* can help make code more understandable without extra comments (see the third instance for a good example, see the first for a bad one). I'm not entirely sure about the compiler specifics but I'm guessing those lambdas just get optimized out and have no performance hit, so I'm not concerned there.

I'm somewhat used to lambdas although not extensively, so the syntax, while somewhat new, isn't too hard for me to figure out. If we want to bring this up at the standup tomorrow @collinsc this would be a good way to ask for some opinions on the issue and maybe add a new point to the coding standard if we come to a decent answer.

src/tfr_sensor/src/light_detection_action_client.cpp
RyanBerge: If the light circuit on the wall doesn't work we'll hit the wall and get disqualified.

I don't think we could conceivably end up in a position where we're backing in the wrong direction, because we don't start that routine unless we see the ArUco marker.

RyanBerge:
I don't think I've ever used Lambdas for anything other than anonymous functor objects to be passed into algorithms (like in the second case).  @AdamuKaapan I don't know what the performance is, but I assume is negligible.

src/tfr_sensor/src/light_detector.cpp
collinsc: @AdamuKaapan I think iterators are more maintainable, than indexing for areas that you don't care about performance.

collinsc: Which we don't in this method, we do in the previous one that processes the video stream

src/tfr_sensor/src/light_detection_action_client.cpp
collinsc: I think this would be such a fatal error that we couldn't/don't want our system to attempt to recover. The user would need to recognize it and do manual control. 

When we integrate this with communication, we can make sure to put a message or status code letting the user know what is going on, but autonomous recovery is probably not desirable.

collinsc:
@AdamuKaapan  @RyanBerge

I totally hear you guys. If that's the way you feel I'll change it.

However before we move away from the newer syntax, let me give you the short sales pitch on how and why you can get concrete benefits from adopting this stuff in your code. I attended a conference a while back, and we talked about this stuff a lot, and I totally got won over. I understand the need for readability, but there are real concrete benefits, to getting used to and adopting these newer constructs.

`foreach` is a member of `std::algorithm`. That library was designed to automate common tasks, and give an extra edge to problem solving, with a diverse array of options. You can get really efficient concurrency, lazy evaluation, and all sorts of fancy searches and sorts, and the whole library is meant to cleanly interact with lambdas, iterators, and the style of programming I demonstrated above. The best part is it is part of the standard, so that means it's portable, and accessible. It's designed for maintainability, flexibility and performance. It has good interoperability with the rest of the stl containers and libraries, as well as threading support. 

`foreach`, provides supports `std::function`, which provides a type safe alternative to function pointers, with minimal overhead, and without sacrificing interoperability with stl containers. lambdas conform with `std::function`, lambdas can be passed in as a direct substitute. (however they do get wrapped in an `std::function` object behind the scenes.) 

As far as lambdas go, I know quite a bit about the subject. I attended a conference where we implemented one from scratch. Basically the compiler generates a functor for you, without you having to type it, and no heap allocation. Captures of the lamda are stored by reference or value as data members of the functor, and one `operator()` overload method is created with your function body. Basically 0 cost. This article does a great job of explaining it: https://blog.feabhas.com/2014/03/demystifying-c-lambdas/.

`std::algorithm` also provides easy integration to the new future/promise c++ concurrency model which started in 14, and is continuing to evolve.

The idea is with the use of std::algorithm you don't have to compromise between compile time type safety, maintainability, portability and performance.

Implementing it more, makes more new language features in the standard accessible, and the whole point of these evolutions, is _not_ to make our code more "pretty" , or follow trends. The reason we do it is to write better c++. That means more type safe, more maintainable, more portable, more performant, and in general to get more fecal matter accomplished in less time.

So if you don't think this use case is performant, that's fine, I'll change it. However, please, I implore you, please read up on this stuff, learn it, breathe it in, get used to seeing it in code, you will write more modern effective c++ because of it. 

Let me know what you guys think.

src/tfr_sensor/src/light_detector.cpp
LinuxKid21: We obviously still need to figure out our stance on the lambdas, but why are we changing from an array to a class for the colors? In my opinion, nothing would be more clear, concise and maintainable than a for loop.
for(int i =0;i < 3; i++)
    stats[i] = intensities[i]/image.rows*image.cols;

collinsc:
@AdamuKaapan I made your requested changes

collinsc:
@RyanBerge I made your changes.

@AdamuKaapan I don't think we know enough about the failure points of our application, at this state of our application. So I don't think that change is appropriate at this time. We also do not have the ability to properly test any autonomous recover options available to us for that edge case at this point in time, so anything we do will be speculation. I think the points of failure will become known on integration testing, and we should address those issues there.

So I did not implement a change.

RyanBerge:
@collinsc Thanks.  It looks like you forgot to actually delete the `scale` lambda (even though it is no longer used).

I think I agree that we don't know how to deal with error recovery right now, so it's fine to leave that out for now.

collinsc:
@RyanBerge I did it as an --amend, let's push this through

RyanBerge (approved):

AdamuKaapan (approved):

PR #21, opened 2018-02-01
Navigation action server integration

collinsc:
# Description
I integrated the navigation action server with the `navigation_stack`'s public interface.

The main navigation stack node is called `move_base`. I wrote the plumbing to integrate that node with our existing sensors. It exposes our dear friend the SimpleActionServer to the user, and accepts PoseStamped messages as it's goal topic. From an integration perspective, the main work was just wrapping the previously written goal setting code to tf-friendly Pose Stamped messages and sending those to the client. That is done in navigator.cpp.

I also refactored the server to stop using message level location constants as per @Jebbs request. I put these in a shared header `location_codes.h` in `tfr_utilities`. I refactored the api to use these exclusively. Unfortunately ros messages do not support user defined types, so I had to use `static_cast` to move the codes into and out of messages.

I also fixed a few parameter bugs and reorganized our hardware sensor platform. This is to prepare for integration testing with sensor fusion on 1/5. That's why there are some random parameters changed here and there in the later commits.

# Not Done
I have not refactored this to include the math to move the goal along the mining line, because we haven't figured out a safe way to do it yet. That should be it's own pull request.

# Testing
Unfortunately this has now been integrated with our sensors. However besides our existing navigation unit tests, and building there isn't much you can do to try it at home.

What I can do is upload a picture of what it does, figure 1 shows the navigation stack guiding the rover down a hallway.
__Figure 1: Close up__
image

You wee the `bin_footprint`  closest to us, representing where the bin is. Slightly in front of it you see the `base_footprint` of the robot. Surrounding the robot is a purple bounding box, a measured replica of the rolling sensor platform.

Extending in front of the rover is a short purple line, representing the specific twist message being sent to the motors.

__Figure 2: Top Down Navigation__
image

Figure 2 shows the same scenario from a larger scale. You see the green global path extending from the robot to the red destination arrow. You will notice we filter out the point cloud at distances > 3m to avoid the noise in the duo image processing. 

The white square is 4x4m and follows the robot and represents what we take into account for local path planning. 

The grey square is 20x20m and follows the robot, and represents how much stored obstacle information we keep in memory before discarding it.


src/tfr_msgs/action/Navigation.action
```diff
@@ -1,22 +1,16 @@
 #goal msg
 Header header
-uint8 location_code
+uint8 location_code
```
Jebbs, line 3: Added whitespace?!

src/tfr_navigation/package.xml
```diff
@@ -4,20 +4,19 @@
   <version>0.0.0</version>
   <description>The tfr_navigation package</description>
 
-  <!-- If you are the actual maintainer of this package, please say so here. -->
-  <maintainer email="ryan.berge@trickfirerobotics.com">Ryan Berge</maintainer>
   <license>BSD</license>
+
+  <maintainer email="ryan.berge@trickfirerobotics.com">Ryan Berge</maintainer>
```
Jebbs, line 9: Why is Ryan the maintainer?

src/tfr_navigation/src/navigation_action_client.cpp
```diff
@@ -14,11 +15,11 @@ int main(int argc, char** argv)
 
     NavigationClient client("navigation_action_server");
     //test preemption
-    client.navigate(tfr_msgs::NavigationGoal::TO_MINING);
-    client.stop_all();
+    client.navigate(tfr_msgs::LocationCode::MINING);
+//    client.stop_all();
```
Jebbs: Commented out code should be removed.

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -59,17 +64,17 @@ move_base_msgs::MoveBaseGoal NavigationGoalManager::initialize_goal() {
  */
 move_base_msgs::MoveBaseGoal NavigationGoalManager::get_updated_mining_goal(geometry_msgs::Pose msg)
 {
-    if (location_code == tfr_msgs::NavigationGoal::TO_MINING){
+    if (goal == tfr_msgs::LocationCode::MINING){
         nav_goal.target_pose.header.stamp =ros::Time::now();
-        //TODO integrate reference frame when bin tf publisher is finished
-        double y_position = msg.position.y; 
-        int sign = (y_position > 0) ? 1 : -1;
+        double v_position = msg.position.y;
```
Jebbs: Back to `v_position`?

src/tfr_navigation/src/navigator.cpp
```diff
@@ -44,84 +48,97 @@ Navigator::Navigator(ros::NodeHandle &n,
  *      -uint8_t code corresponding to the finsal status described in 
  *      Navigation.action in the tfr_msgs package
  *      -Pose describing our final position
- *  NOTE NOT THREAD SAFE
+ *  NOTE careful with the shared pointers in this class if threading becomes
```
Jebbs, line 46: Is this listed somewhere not in code in case it comes up?

Jebbs (changes requested):
I'm pretty much fine with all of this, but I don't think the `LocationCode` enum class should be in `tfr_msgs`. I think it belongs in `tfr_navigation` since that's the place its actually used. 

Edit:
It looks like you meant to put it in `tfr_utilities`, but accidentally put it in `tfr_msgs`. I would be OK with it being in `tfr_utilities`, but I still think that `tfr_navigation` is a better place.

src/tfr_msgs/action/Navigation.action
collinsc: yeah that's weird, I don't know why that happened.

src/tfr_navigation/package.xml
collinsc: It just got auto generated when ryan set up the repository. There is a card in the trello asking us to put some general trickfire email down at some point in all of our packages.

src/tfr_navigation/src/navigation_action_client.cpp
collinsc: I don't agree with you in this context, It's a test class, that is on the chopping block to be removed in system integration. Next week my job is to literally throw out this code and incorperate the major responsibilities into executive.

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: odd I don't know why that change didn't make it.

src/tfr_navigation/src/navigator.cpp
collinsc: No not really, I couldn't think of a place that would be more useful to put it than the header of the class if it eventually becomes a priority.

collinsc: If you have an idea, I am all ears.

collinsc:
@Jebbs I don't agree with your requested change to putting it to navigation. It's a message level constant that will be shared between executive and navigation. I'll move it to utilities if that's where you want it, but I won't move it to navigation, that doesn't sound right to me.

collinsc:
@Jebbs I applied your changes! Thanks for the review.

src/tfr_navigation/package.xml
RyanBerge: It actually wasn't auto-generated; I just figured it was better having some email than no email.

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -19,37 +20,41 @@ NavigationGoalManager::NavigationGoalManager(const GeometryConstraints &c,
...
         default:
-            ROS_WARN("location_code %d not recognized", location_code);
+            ROS_WARN("location_code %u not recognized",
+                    static_cast<uint8_t>(goal));
     }
+    //set relative rotation (none)
+    nav_goal.target_pose.pose.orientation.x = 0;
+    nav_goal.target_pose.pose.orientation.y = 0;
+    nav_goal.target_pose.pose.orientation.z = 0;
+    nav_goal.target_pose.pose.orientation.w = 1;
```
RyanBerge, line 56: What is "relative rotation" here?

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -59,17 +64,17 @@ move_base_msgs::MoveBaseGoal NavigationGoalManager::initialize_goal() {
...
 move_base_msgs::MoveBaseGoal NavigationGoalManager::get_updated_mining_goal(geometry_msgs::Pose msg)
 {
-    if (location_code == tfr_msgs::NavigationGoal::TO_MINING){
+    if (goal == tfr_utilities::LocationCode::MINING){
         nav_goal.target_pose.header.stamp =ros::Time::now();
-        //TODO integrate reference frame when bin tf publisher is finished
         double y_position = msg.position.y; 
         int sign = (y_position > 0) ? 1 : -1;
         nav_goal.target_pose.pose.position.y = sign*std::min(
                 constraints.get_mining_line_length()/2, std::abs(y_position));
```
RyanBerge: Is the change for the mining length in a newer pull request?

src/tfr_navigation/src/navigator.cpp
RyanBerge: I think it's important in a more general way for all of us to remember that _any_ form of node-to-node communication is inter-process communication and therefore can run into shared data problems.  Callback functions for Services and Actions are generally running in different processes than the data they're interacting with.

RyanBerge (changes requested):
(this isn't a request for changes so much as it is a request for answers to the couple questions I had)

src/tfr_navigation/package.xml
collinsc: I think catkin will also do a build warning if you don't have this tag filled in anyway.

src/tfr_navigation/src/navigator.cpp
collinsc: Yeah the way I am using the const shared pointers to const objects right now is completely thread safe, but if someone get's careless in the future and starts doing something silly, they need to be aware they are in a threaded enviroment and really watch their shared memory.

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: (0,0,0,1) means no rotation. 
image

collinsc: This site is also really handy for making quaternions human readable

http://quaternions.online/

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: Yup #25

src/tfr_navigation/src/navigator.cpp
collinsc: However that is a side effect of the action server in general, only if you are using the execute  callaback version of the object. If you run it using the goal callback, then it runs on the main thread of the node. See this documentation for details.http://docs.ros.org/jade/api/actionlib/html/classactionlib_1_1SimpleActionServer.html

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: the relevant frame is `bin_footprint` which is a point in the dead center of the bin, at ground level.

so the instruction is go `mining_distance` east of `bin_footprint` with no rotation.

collinsc:
@Jebbs @RyanBerge Thanks for asking those concurrency questions, I did a pretty thorough double check in the documentation and spotted a concurrency error.

So sanity check me, but on the user thread I have a `const shared ptr`, continuously updating it's reference to an object in the main thread.

In the execute thread, when I need to access it, I create a local `const shared ptr` to whatever main thread `const shared ptr` is pointing to at the time, which is guaranteed to be atomic by boost. (i would normally have liked to use the `std::shared_ptr`, but ros uses boost) 

Then I do nonatomic access on the local `const shared ptr`. 

The rest of the operations of the class do not have any shared memory.

My patch is in the most recent commit.

collinsc:
Which is failing to build in executive, huh.

I didn't make any changes that even touched the build. I don't think this jenkins failure is related to my most recent commit. 

I think we can probably ignore it because it's randomly failing on non-production code.

```
 /var/lib/jenkins/workspace/Jenkins Build - NasaRmc2018/src/tfr_executive/src/executive_echo.cpp: In function ‘int main(int, char**)’:
/var/lib/jenkins/workspace/Jenkins Build - NasaRmc2018/src/tfr_executive/src/executive_echo.cpp:43:46: error: ‘tfr_msgs’ was not declared in this scope
     ros::Publisher chatter_pub = n.advertise<tfr_msgs::Echo>("executive_echo", 1000);
                                              ^
/var/lib/jenkins/workspace/Jenkins Build - NasaRmc2018/src/tfr_executive/src/executive_echo.cpp:43:36: error: parse error in template argument list
     ros::Publisher chatter_pub = n.advertise<tfr_msgs::Echo>("executive_echo", 1000);
                                    ^
/var/lib/jenkins/workspace/Jenkins Build - NasaRmc2018/src/tfr_executive/src/executive_echo.cpp:43:84: error: no matching function for call to ‘ros::NodeHandle::advertise(const char [15], int)’
     ros::Publisher chatter_pub = n.advertise<tfr_msgs::Echo>("executive_echo", 1000);
```

src/tfr_navigation/src/navigator.cpp
RyanBerge: I'm not sure what you mean.  Can you expand on that?

src/tfr_navigation/src/navigation_goal_manager.cpp
RyanBerge: So the relative rotation is relative to the orientation of the bin?

src/tfr_navigation/include/tfr_navigation/navigator.h
```diff
@@ -34,17 +36,24 @@ class Navigator
...
         Server server;                      //handle to as
+        //NOTE delegate initialization of server to ctor
+        NavStack nav_stack;
         //NOTE delegate initialization of subscriber to ctor
         ros::Subscriber odom_subscriber;    
-        //NOTE delegate initialization of mgr to ctor
-        NavigationGoalManager goal_manager;
 
         //needed msgs
         tfr_msgs::NavigationFeedback feedback{};
```
RyanBerge: As far as I can tell, you don't need to have `feedback` as a member variable.  You can just have `update_feedback` return a fresh object, since it's building it from scratch anyway.

This is a little better, I think, and it has the added benefit of removing any possibility that future edits to this class could run into race conditions if they were for some reason modifying this value.

RyanBerge: Same thing goes for the `result` message, I believe.

RyanBerge:
@collinsc Actually, I believe there _is_ a shared data problem in your code.  The example that you link in your comments from `update_position` is a copy from a `const sensor_msgs::ImageConstPtr` to a `sensor_msgs::ImageConstPtr` (two ConstPtrs).  Your code, however, does a _dereference_ from a `nav_msgs::OdometryConstPtr` and assigns it to a member variable with a type of `nav_msgs::Odometry`.  I don't have the generated source code for the Odometry type, but that is not a pointer assignment, and I don't think we can assume they've overloaded that operator with thread-safe code.

RyanBerge:
I hit the wrong button.  I didn't mean to close the pull request.

src/tfr_navigation/src/navigator.cpp
collinsc: When you construct an action server you have a lot of different constructor choices, but there are two main categories:
1. _Use Execute Callback_: If you use the execute callback, it spins up automatically in a separate thread. This allows you to make blocking calls. It get's called when you send it a goal. Setting the execute callback disables the goal callback. 

2. _Use the Goal Callback_: If you use the goal callback, it does not spin up a thread, your callbacks will run on the main thread in the context of the parent node. The goal callback get's called,  when you set a goal, you can also use a preempt callback in this context. This example in the tutorials shows these two used in conjunction with a subscriber callback to take an average of some sensor data. http://wiki.ros.org/actionlib_tutorials/Tutorials/SimpleActionServer%28GoalCallbackMethod%29.

src/tfr_navigation/include/tfr_navigation/navigator.h
collinsc: Will do!

collinsc:
@RyanBerge Let me take a look!

collinsc:
@RyanBerge You are right that is unsafe. That's weird could have sworn I changed that to a ConstPointer a couple of commits ago. Either I did a bad rebase, or I forgot to implement that change.

src/tfr_navigation/include/tfr_navigation/navigator.h
collinsc: Well actually looking at the code, after the new thread safety changes, I don't think we have any more race conditions. 

And, I don't think making that change provides good interface consistency with the current position method.

So either we leave it alone or we do a full scale refactor of the interface, and have update position return a different object.

collinsc: If that's the scope of the change you are suggesting that needs to be a seperate pull request.

collinsc:
I amended the last commit, for that pointer ommision. @RyanBerge See the question I asked and let me know if that is something we need to do.

src/tfr_navigation/src/navigation_client.cpp
```diff
@@ -65,5 +66,5 @@ void NavigationClient::feedback(const tfr_msgs::NavigationFeedbackConstPtr& feed
 void NavigationClient::finished(const actionlib::SimpleClientGoalState &state, const
         tfr_msgs::NavigationResultConstPtr &result)
 {
-    ROS_INFO("result recieved, code: %d", result->status);
+    ROS_INFO("result recieved");
```
AdamuKaapan: Typo: *received

src/tfr_navigation/src/navigation_client.cpp
```diff
@@ -56,7 +57,7 @@ actionlib::SimpleClientGoalState NavigationClient::get_state()
  * */
 void NavigationClient::feedback(const tfr_msgs::NavigationFeedbackConstPtr& feedback)
 {
-    ROS_INFO("feedback recieved, code: %d", feedback->status);
+    ROS_INFO("feedback recieved");
```
AdamuKaapan: Typo: *received

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -19,37 +20,41 @@ NavigationGoalManager::NavigationGoalManager(const GeometryConstraints &c,
         ROS_WARN("    finish_line: %f",
```
AdamuKaapan, line 19: Is this tab intentional?

src/tfr_navigation/src/navigation_goal_manager.cpp
```diff
@@ -59,17 +64,17 @@ move_base_msgs::MoveBaseGoal NavigationGoalManager::initialize_goal() {
  */
 move_base_msgs::MoveBaseGoal NavigationGoalManager::get_updated_mining_goal(geometry_msgs::Pose msg)
 {
-    if (location_code == tfr_msgs::NavigationGoal::TO_MINING){
+    if (goal == tfr_utilities::LocationCode::MINING){
         nav_goal.target_pose.header.stamp =ros::Time::now();
```
AdamuKaapan: Formatting nitpick: add a space after the equals sign

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: yes

src/tfr_navigation/src/navigation_client.cpp
collinsc: eye kan spel

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: got it

src/tfr_navigation/src/navigation_goal_manager.cpp
collinsc: yes.

RyanBerge (dismissed):

AdamuKaapan (approved):

RyanBerge (approved):

collinsc:
I did jeremy's changes a while ago to the letter, and I have two approvals, so I am going to merge this.

PR #22, opened 2018-02-01
Bin controller

RyanBerge:
This pull requests introduces an intermediary interface between our primary systems and the hardware layer for controlling the tilt of the bin.  It is implemented as an Action Server with a name of `bin_control`.  I can be called with a goal to either raise or lower the bin, and will return when the bin has reached the desired position.

This pull request does not introduce a client component for testing, so this code is mostly untested.


collinsc:
@RyanBerge  Can you provide a brief pull request description telling me what you did and how you want me to test your code?

src/tfr_control/src/bin_control_server.cpp
```diff
@@ -0,0 +1,82 @@
+#include "bin_control_server.h"
+#include <boost/bind.hpp>
+#include "std_msgs/Float64.h"
+
+
+// I could also pass in the name of the server, but it would end up being a literal
+// string anyway unless it also gets passed into the node, which seems a bit excessive.
+// If it's going to be a literal string we might as well put it as close to its destination
+// as possible so it's clear what it's being used for.
```
collinsc, line 15: Literal strings are fine for things that shouldn't change, or we don't ever really see changing.

src/tfr_control/src/bin_control_server.cpp
```diff
@@ -0,0 +1,82 @@
...
+            "bin_position_controller/command", 100);
+    server.start();
+}
+
+// Action Server callback function
+void BinControlServer::ControlBin(const tfr_msgs::BinGoalConstPtr& goal)
+{
+    std_msgs::Float64 command;
+    tfr_msgs::BinResult result;
+
```
collinsc, line 30: There needs to be a preemption check at the start of every action server execute cb.

src/tfr_control/src/bin_control_server.cpp
```diff
@@ -0,0 +1,82 @@
...
+    if (goal->command_code == goal->RAISE_BIN)
+    {
+        result.return_code = result.BIN_RAISED;
+    }
+    else if (goal->command_code == goal->LOWER_BIN)
+    {
+        result.return_code = result.BIN_LOWERED;
+    }
+
+    server.setSucceeded(result);
```
collinsc, line 61: Is there any reason this could fail? If so how could it fail? Would you ever need to use `setAborted`?

src/tfr_control/include/tfr_control/bin_control_server.h
```diff
@@ -0,0 +1,39 @@
...
+#define BIN_CONTROL_SERVER_H
+
+#include <ros/ros.h>
+#include <actionlib/server/simple_action_server.h>
+#include <tfr_msgs/BinAction.h>
+#include <mutex>
+#include <condition_variable>
+
+using Server = actionlib::SimpleActionServer<tfr_msgs::BinAction>;
+
```
collinsc, line 24: Give me some headers describing what going on in this serve, also describe explictly what it publishes and where in the header.

src/tfr_control/src/controller.cpp
```diff
@@ -30,20 +30,33 @@ namespace tfr_control
         registerInterface(&joint_effort_interface);
     }
 
-    void Controller::read() {
+    // Return true if bin is finished moving to its new position.
+    // TODO: A more explicit return type might be better here, or a possible redesign.
+    //       If other values read end up needing to be returned, then this should 
+    //       probably return some data structure. It may be that read values get published
+    //       out on topics instead.
+    bool Controller::read()
```
collinsc, line 38: Let's make this method name a little more specific. If this is supposed to be the controller for all of the hardware interfaces on the whole robot, then methods just like `read` or `write` are not descriptive enough. Maybe `read_bin` or `write_bin` might improve maintainability.

src/tfr_control/src/bin_control_server.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void BinControlServer::SignalBinController()
```
collinsc: This synchronization callback needs a comment header describing who calls it and why.

src/tfr_control/src/controller.cpp
```diff
@@ -30,20 +30,33 @@ namespace tfr_control
...
+    bool Controller::read() 
+    {
         // TODO: Waiting on further hardware development to implement
+
+        // Return true if bin has finished moving to its new position.
+        return false;
     }
 
-    void Controller::write() {
+    void Controller::write()
```
collinsc: see above comment

src/tfr_control/src/bin_control_server.cpp
```diff
@@ -0,0 +1,82 @@
...
+        //       polling loop here, but I don't think there's an easy way to tell the hardware_interface
+        //       to simply stop moving the bin; it takes in a target angle and simply goes until it gets 
+        //       there. If we think supporting preemption is important for this component, we can 
+        //       investigate further options.
+
+        // http://en.cppreference.com/w/cpp/thread/condition_variable 
+        // https://stackoverflow.com/questions/16350473/why-do-i-need-stdcondition-variable
+        // wait() will release the lock. Sometimes the signal will be notified by accident, so
+        // a check on bin_task_completed is required. wait() reaquires the lock after it unblocks.
+        signal.wait(signal_lock);
```
collinsc: From what I understand from you code, 
1. client sets goal
2.  server receives it and publishes on topic
3.  block until completed.
  - controller.cpp sets condition variable when it finishes. 
4.  return result.

Thats fine, but it took me too long to get there from reading your code. 

You need to take a few steps to make this more readable and maintainable for future readers:
- Extract all the blocking synchronization code into a separate method: e.g. `wait_for_bin()`
  - In that method tell the readers what's happening, who we are blocking on, and what events trigger this.
  - I think at least 10 minutes of reading could be shaved of "reader comprehension time" if we did this, synchronization usually needs a little extra care in documentation if you want to maintain it..
- If we aren't supporting preemption, explain why in the class header, and in the description of the method.

src/tfr_control/src/controller.cpp
collinsc: Follow up question is `controller.cpp` the only controller class for the whole robot?

collinsc (changes requested):
Most of the code looks good, I asked for some readability changes.

src/tfr_control/src/bin_control_server.cpp
collinsc: even if we don't support it during the blocking call.

collinsc:
@RyanBerge I have a change an additional that I want to request, can you refactor the action server to just successfuly exit and set all of the correct messsages. I can integrate the dumping action server with this next sprint if you do.

src/tfr_control/include/tfr_control/bin_control_server.h
RyanBerge: Sure.

src/tfr_control/src/bin_control_server.cpp
RyanBerge: What's the reason for this?  I can't imagine a situation where we would immediately preempt an action after requesting it.  Is the preemption flag also used to detect weird internal ROS failures?

src/tfr_control/src/bin_control_server.cpp
RyanBerge: Sure.  Yeah, you've got it right.

src/tfr_control/src/bin_control_server.cpp
RyanBerge: In this current implementation it uses `setAborted` only if the goal doesn't use one of the predefined inputs (`RAISE_BIN` or `LOWER_BIN`).

Conceivably, the raising of the bin _could_ fail (though I really hope this wouldn't happen), but we don't have a mechanism in place yet to detect that and I didn't write the signaling interface in such a way that it can tell us that, either.  I figured that would be up to a future edit.

src/tfr_control/src/bin_control_server.cpp
RyanBerge: Sure.

src/tfr_control/src/controller.cpp
RyanBerge: `controller.cpp` is the class we use as an interface between the hardware and our message-passing system, though that interface is also shared by a ROS class called `controller_manager` (both of these classes, as well as my bin control server, live in the `controller_launcher` node).  So yes; `controller.cpp` is for the entire robot.  `read()` reads from our encoders (all of them) and does some as-yet-undefined action with the data.  `write()` writes voltage values out to the I2C connection for all our actuators. 

This was all set up by @AdamuKaapan, so he can correct me if I've gotten anything wrong.

RyanBerge:
@collinsc What do you mean by "just successfully exit"?  Are you saying do placeholder code so it doesn't block for now?

src/tfr_control/src/bin_control_server.cpp
collinsc: It's just a best practice, I haven't seen a tutorial without it. It's just one of those things we should automatically do.

src/tfr_control/src/bin_control_server.cpp
collinsc: That was more of my question. If there is no way to detect failure then we don't have to check, but if the provided interface provides a way to check for failure we should do it here.

collinsc:
@RyanBerge Yep that way I can integrate it with the dumping server. Thanks!

collinsc (dismissed):

collinsc:
Looks good! I approved the updates, thanks for doing them. It would be great if we had a short description in the pull request header for posterity.

src/tfr_control/src/controller.cpp
AdamuKaapan: Yeah, that description looks correct. I think that we might want to find a different way to pass out the return value, whether we save them in some other local variable, put them out on topics, or something else @RyanBerge . While I can't *think* of anything else that might need a return value from this method right now, it still feels like a kinda kludgey design and would prefer to find a better solution. And @collinsc due to all of the sensor inputs coming from the same hardware device it makes it much simpler/resource efficient to put it in one method, as much as I want to split it apart (although we could perform separate processing of said inputs on a separate input).

src/tfr_control/src/controller.cpp
```diff
@@ -61,7 +76,8 @@ namespace tfr_control
         joint_effort_interface.registerHandle(handle);
     }
 
-    double Controller::get_update_time() {
+    double Controller::get_update_time() 
+    {
```
AdamuKaapan, line 80: :+1: thanks for fixing the curly brace alignment that I screwed up

src/tfr_control/src/controller.cpp
AdamuKaapan: Although now that I've seen the rest and seen that the bin control server is integrated with the controller_launcher itself, this isn't quiiiite as bad. I think I'd still prefer a slightly better solution though.

AdamuKaapan (changes requested):
Only request is to investigate into possible alternate solutions to having the `read()` method in `controller` returning a bool. If you come back to the conclusion that this is the best solution then that's alright though.

RyanBerge:
@AdamuKaapan I'm definitely not convinced this is the best solution.  My plan was to refactor it once I knew what the design in `read()` and `write()` was going to look like (and writing that hardware interface is my next task).  Do you want me to change it now?

collinsc (approved):
Thanks for adding the description! I am reapproving after the rebase.

AdamuKaapan:
@RyanBerge if you have plans to change it then that's totally fine, I'll go ahead and approve it for now.

AdamuKaapan (approved):

PR #23, opened 2018-02-02
Dumping Action Server

collinsc:
# Prerequisites
This branch expands on #20 so that branch should be reviewed and merged before this one is tackled.

# Description
I wrote the dumping action server and integrated it with light detection. It takes an empty action message as a goal. 

It required a few things I needed to make:
- __Camera__: I needed a camera to test with so I integrated one of the 1mp 1080p robotics cameras, and calibrated it using the cv_camera node. There is a rear_cam.launch file that reads in my calibration parameters and spins up the camera.
- __On Demand Image Retrieval__: I needed on demand image retrieval from the rear camera stream. I made an image wrapper node that accomplishes this task. It does so by wrapping a topic and storing it as a super efficient, shared pointer, and retrieving that value on request as a ros service. I instantiate it in a rear_cam_wrapper.launch specifically for the purpose of wrapping the rear camera stream. 
- __Fake Aruco Output__: I needed a message to work with to simulate the messages aruco would give me, so I faked one, `Aruco.action` 

The algorithm takes a couple of parameters, see following sections to see their purposse.
- `min linear velocity`: our slow backing speed
- `max linear velocity`: our normal backing speed
- `min angular velocity`: our slow turning speed
- `max angular velocity`: our normal backing speed
- `image service`: the image service

As a precondition it checks to see if we can see the board, if it can't it rejects the goal. (This goes along with a few conversations we had as a team a long time ago, if we need to change it, super easy one line fix).

It starts the light detection mechanism and does the following algorithm until it see's the light:
1. Check for light
  1.1 if detected stop motors & exit
2. Check for preemption
  2.1 If false stop motors & exit
3. Check for Aruco using image service to get most recent data.
  3.1 if we can see aruco back towards the bin at full speed, turning towards the center
  3.2 else back up at minimum speed at current heading
4. rinse. lather. repeat.

# TODO
This relies on aruco, once that is online we need to integrate the aruco action server.

# Testing
This relies on our hardware and cameras, so this again will be pretty hard to test at home. However I can show you how I tested on the jetson. 

__Figure 1: Execution Output__
screenshot from 2018-02-01 18-16-45

On the left terminal in figure 1, you see the output of the action server.
It connects, and accepts a goal, stops and exits with a success status. This exit is triggered by the light detection device.

On the right you see it output of twist msgs on `/cmd_vel`.  As a surrogate for aruco, in the main control loop I am giving it a fake hard coded aruco action server result. You will notice the last packet recieved is and all 0 stop message.

There is no human perceptible latency for this reaction, so mission accomplished.


src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -0,0 +1,164 @@
...
+ */
+void Dumper::dump(const tfr_msgs::EmptyGoalConstPtr &goal) 
+{  
+    //check to make sure we can see the board
+    auto initial_estimate = get_aruco_estimate();
+    if (initial_estimate.markers_detected == 0)
+    {
+        server.setAborted();
+        return;
+    }
```
RyanBerge, line 49: `dump` is being called by Executive, right?  Do we have a plan in place for what to do if we can't see the board here?

It might be good to use a non-empty goal that has some error codes as a return value so that Exec can attempt to take some appropriate action.  Alternatively, I can see an argument that says the Dumping system needs to try and handle this problem somehow.

src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -0,0 +1,164 @@
...
+        if (estimate.markers_detected == 0)
+            move_blind();
+        else
+        {
+            update_control_msg(estimate);
+            velocity_publisher.publish(cmd);
+        }
+    }
+    stop_moving();
+    server.setSucceeded();
```
RyanBerge: Presumably this is where we'd add the call to the BinController to actually dump?  If so, let's add a `TODO` comment.

RyanBerge (changes requested):
Just had a couple small comments/requests.  I also want to wait until the Aruco PR gets merged to approve this one, for obvious reasons.

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: Sure will do!

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: You know we haven't talked about it. From what I can think of there are 3 scenarios.

1. We are pointing in the wrong direction and have room to turn: If this is the case navigation did something really wrong, and we need to run localization again. Since executive ran localization the first time, I think executive should do it again.

2. We are pointing the right direction and we are too close to see the marker: then we should not try and localize, because we will hit the wall and just back up blind. 

3. We are pointing the wrong direction and too close: at this point we are just pretty screwed, we can't turn effectively, and don't know where we are, we should have manual control take over.

It's kind of hard to write code to cover all 3 scenarios well, if we end up any of them our positional information might be compromised or untrustworthy, so it would be difficult to determine a clean solution with our current sensor sources. 

What I suggest we can do is make one of the scenarios more likely than the other, depending on where we set the finish line, and optimize for one or the other using the setup we have.

I think it's a little less risky from a collision standpoint to try and run dumping as far away from the obstacle zone, and as close to the bin as possible. The way we have set up dumping and planned it isn't obstacle aware, and might take a really long time to make it obstacle aware.

If we choose to use fiducial odom publisher we could get some pretty good localization data when we are close we have some more tools at our disposal when we can see the bin. We also have another one of those hd  robotics cams, so that would be pretty easy to incorporate that forward facing.

Regardless, Let's talk about it at the meeting and make it next pull request.

RyanBerge: All three of those scenarios seem pretty unlikely.  If navigation fails enough that we're in this situation, it may be best to switch to tele-op.  I'm not sure.

collinsc:
@RyanBerge I put in that comment. I think any error handling for the dumping action server requires a little team discussion, and might be out of the scope of this request. So let's address it as a team.

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: @RyanBerge Hey we forgot to talk about this at the meeting. 

Here is what I propose. Let's make a maintenence comment here to address this in testing. However I don't think we know enough to anything here yet.

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+#include <actionlib/server/simple_action_server.h>
+#include <actionlib/client/simple_action_client.h>
+/*
+ * The dumping action server, it backs up the rover into the navigational aid
+ * slowly.
+ *
+ * Its first step is to make sure it can see the aruco board, it will abort the
+ * mission if it can't. 
+ *
+ * I backs up at a set speed until it get's really close and loses sight of the
```
AdamuKaapan: Type to fix later: *It (doesn't have to be done in this PR, this is super not crucial).

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+                const DumpingConstraints &constraints);
+        ~Dumper() = default;
+        Dumper(const Dumper&) = delete;
+        Dumper& operator=(const Dumper&) = delete;
+        Dumper(Dumper&&) = delete;
+        Dumper& operator=(Dumper&&) = delete;
+
+    private:
+        void dump(const tfr_msgs::EmptyGoalConstPtr &goal);
+        tfr_msgs::ArucoIntegrateResult get_aruco_estimate();
```
AdamuKaapan: Current coding standard is `lowerCamelCase` for all of these

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+        void move_blind();
```
AdamuKaapan: Current coding standard is `lowerCamelCase`

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+        void stop_moving();
+        void update_control_msg(const tfr_msgs::ArucoIntegrateResult &estimate);
+
```
AdamuKaapan: Current coding standard is `lowerCamelCase`

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+        Dumper(const Dumper&) = delete;
+        Dumper& operator=(const Dumper&) = delete;
+        Dumper(Dumper&&) = delete;
+        Dumper& operator=(Dumper&&) = delete;
+
+    private:
+        void dump(const tfr_msgs::EmptyGoalConstPtr &goal);
+        tfr_msgs::ArucoIntegrateResult get_aruco_estimate();
+        void move_blind();
+        void stop_moving();
```
AdamuKaapan: Current coding standard is `lowerCamelCase`

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+        struct DumpingConstraints
+        {
+            private:
+                double min_lin_vel, max_lin_vel, min_ang_vel, max_ang_vel;
+            public:
+                DumpingConstraints(double min_lin, double max_lin, 
+                        double min_ang, double max_ang):
+                    min_lin_vel(min_lin), max_lin_vel(max_lin),
+                    min_ang_vel(min_ang), max_ang_vel(max_ang) {}
+                double get_min_lin_vel() const {return min_lin_vel;}
```
AdamuKaapan: Current coding standard is `lowerCamelCase` for all of these

src/tfr_sensor/src/image_topic_wrapper.cpp
```diff
@@ -0,0 +1,77 @@
...
+        ~ImageWrapper() = default;
+        ImageWrapper(const ImageWrapper&) = delete;
+        ImageWrapper& operator=(const ImageWrapper&) = delete;
+        ImageWrapper(ImageWrapper&&) = delete;
+        ImageWrapper& operator=(ImageWrapper&&) = delete;
+
+    private:
+
+        //subscription callback
+        void set_current(const sensor_msgs::ImageConstPtr &i)
```
AdamuKaapan, line 46: Current coding standard is `lowerCamelCase` for these methods

AdamuKaapan (changes requested):
All looks good to me, just found some naming convention issues.

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
```diff
@@ -0,0 +1,74 @@
...
+
+        //TODO add the Dumping Controller here
+        //TODO add aruco here
+        
+        ros::ServiceClient image_client;
+        ros::Publisher velocity_publisher;
+
+        const DumpingConstraints &constraints; 
+
+        geometry_msgs::Twist cmd{};
```
RyanBerge: Let's go ahead and fold this into `update_control_msg` like we did with the other PR.

src/tfr_dumping/src/dumping_action_client.cpp
```diff
@@ -0,0 +1,25 @@
...
+int main(int argc, char** argv)
+{
+    ros::init(argc, argv, "simple_action_client");
+    ros::NodeHandle n;
+    actionlib::SimpleActionClient<tfr_msgs::EmptyAction> client("dump", true);
+    ROS_INFO("client waiting for server");
+    client.waitForServer();
+    ROS_INFO("client connected to server");
+
+    ros::Duration(1).sleep();
```
RyanBerge: Why are we sleeping here?

RyanBerge (changes requested):
One small change and one question in addition to @AdamuKaapan's requests.

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
collinsc: Oh sure I got those mixed up, I'll fix that.

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
collinsc: yup yup yup.

src/tfr_sensor/src/image_topic_wrapper.cpp
collinsc: yup

src/tfr_dumping/include/tfr_dumping/dumping_action_server.h
collinsc: sure,

src/tfr_dumping/src/dumping_action_client.cpp
collinsc: Probably left over from an old manual test case, I'll delete it.

collinsc:
@AdamuKaapan @RyanBerge I incorperated your changes.

src/tfr_dumping/src/dumping_action_client.cpp
collinsc: I checked it was just leftover test code.

collinsc: /removed it

RyanBerge (approved):

AdamuKaapan (approved):

PR #25, opened 2018-02-02
Nav mining point

collinsc:
# Description
As with the proposed requirements change to move away from the the mining line and get to the mining point. See the pull request on documentation to do so. This pull request implements that change.

# Prerequisite
This code is an additive change to #21, and we need to discuss it as a team so we can't finish it until we do both of those things.

# Testing 
This was tested and verified in the same way as #21, so unfortunately you cannot do any testing at home beyond building and re running the unit tests.

### DIFF
--- a/src/tfr_navigation/src/navigator.cpp
+++ b/src/tfr_navigation/src/navigator.cpp
@@ -6,7 +6,9 @@
  * */
 Navigator::Navigator(ros::NodeHandle &n,
         const NavigationGoalManager::GeometryConstraints &constraints, 
-        const std::string &name, const std::string &bin_frame) : node{n}, 
+        const std::string &name, 
+        const std::string &bin_frame) : 
+        node{n}, 
         goal_manager(bin_frame,constraints),
         server{n, name, boost::bind(&Navigator::navigate, this, _1) ,false}, 
         nav_stack{"move_base", true}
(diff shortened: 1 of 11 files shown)


RyanBerge (changes requested):
I'm gonna wait on a review for this PR until a couple other ones get merged, because there are 36 files changed but most of them are changes I already reviewed and I don't want to parse out the new ones.  Once the other two PRs are done and this is rebased the diff will be a lot easier to read.

collinsc:
Yeah It's only the last commit that is relevant, but I agree, if I have to change the other branch I would need to rebase the changes I made here anyway.

src/tfr_navigation/src/navigator.cpp
```diff
@@ -44,84 +45,97 @@ Navigator::Navigator(ros::NodeHandle &n,
...
-        result.status = tfr_msgs::NavigationResult::OK;
+        nav_stack.cancelAllGoals();
         server.setAborted(result);
     }
     ROS_INFO("Navigation server finished");
 }
 
 
+/*
+ * Prepare a feedback message for sending
```
AdamuKaapan: Wrong comment? Should be "prepare a result message for sending"?

AdamuKaapan (dismissed):
Left a small comment about a comment, can be fixed at a later PR.

collinsc:
@AdamuKaapan I fixed that comment, in https://github.com/TrickfireRobotics/NasaRmc2018/pull/25/commits/be592cc42b2573602a78b06a4976dc04145e0483  for some reason your comment did not get carried over when I rebased.

RyanBerge (approved):

PR #26, opened 2018-02-02
added the Aruco Action Server

LinuxKid21:
# Purpose

To integrate the openCV Aruco library as an action server in ROS

# What Got finished and what needs to be done still

The server is completely finished as is an example client. Right now there is no way to generate aruco boards and the position of the markers is done manually in the code(which is painfully error prone). Aruco also has quite a few parameters, but I did not create a way to change those on the fly because it didn't seem necessary (in fact, the defaults seem to work the best anyway).

I need to create automated tests as well

# Testing

You will need to print out the image file in the tfr_aruco folder (each marker will be 3.2 cm). I find it helpful to mark the top left corner when its done by putting an arrow to it.
You will need a webcam as well.
If you can't print or don't have a webcam then I will be happy to send you recorded messages(as soon as I figure out how to do that)
After building the program, run the following:
roslaunch tfr_aruco test_aruco.launch


src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+    
+    
+        
+        cv::Vec3d boardRotVec, boardTransVec;
+        int markersDetected = cv::aruco::estimatePoseBoard(markerCorners, markerIds, board, cameraMatrix, distCoeffs, boardRotVec, boardTransVec);
+        
+        tfr_msgs::ArucoResult result;
+        result.number_found = markersDetected;
+        result.position = std::vector<float>(3, 0); // 3 members initialized to 0
+        result.rotation = std::vector<float>(3, 0); // 3 members initialized to 0
```
collinsc, line 117: You can eliminate the zero inititialization step by just doing this, not required, but cleaner.
 `tfr_msgs::ArucoResult result{};`
`//... fill result with stuff`

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+        if(markersDetected != 0) 
+        {
+            result.position[0] = boardTransVec[0];
+            result.position[1] = boardTransVec[1];
+            result.position[2] = boardTransVec[2];
```
collinsc, line 122: I am pretty sure there is a copy ctor for:
`result.position = boardTransVec;`

src/tfr_aruco/src/tfr_aruco_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+    
+            result.rotation[0] = boardRotVec[0];
+            result.rotation[1] = boardRotVec[1];
+            result.rotation[2] = boardRotVec[2];
```
collinsc, line 126: see above

collinsc (approved):
I left a couple of suggestions for cleaner ros message handling, but it's not explicitly required. In a future refactor we can integrate it with the jetson hardware. Other than that thanks for cleaning it up!

Jebbs (approved):
No comments from me on top of what @collinsc  already mentioned.

Let's do it.

PR #31, opened 2018-02-09
Added system codes

Jebbs:
### About:
This pull adds enums describing the sub systems and their status codes, where the status codes are any status the system could be in. It also adds functions that parse and generate the system messages.

The idea here is for each sub system maintainer to add a new status to the `StatusCode` enum class and to then add an accompanying string message in that system's parse function.

The status code width was changed from 32 to 16 bits, where the higher order byte represents the sub system relaying the message and the lower order byte represents a status for that sub system. This gives each sub system the ability to have 256 different statuses.

I've also removed the original utilities placeholder, and then updated `mining_echo.cpp` which relied on it. This shows what using the status message and status code in a sub system would be like. It might be a good idea to add a helper function for this to avoid needed to do the cast inline.

### Testing:
There is a simple test that was added to the Utilities package just to show that the system will switch correctly. I didn't feel the need to add a test for all cases.

### Notes:
This design would work, but I'm not entirely happy with it. Feedback would be appreciated.

### DIFF
--- a/src/tfr_msgs/msg/SystemStatus.msg
+++ b/src/tfr_msgs/msg/SystemStatus.msg
@@ -1,3 +1,3 @@
 time time_stamp
-int32 status_code
+uint16 status_code
 float32 data
--- a/src/tfr_utilities/CMakeLists.txt
+++ b/src/tfr_utilities/CMakeLists.txt
@@ -51,6 +51,10 @@ include_directories(
   ${GTEST_INCLUDE_DIRS}
 )
 
+add_library(${PROJECT_NAME}
+  src/status_code.cpp
+)
+
 add_library(tf_manipulator
     ./src/tf_manipulator.cpp
 )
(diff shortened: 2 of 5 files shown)


RyanBerge (dismissed):
This looks good to me.

collinsc (changes requested):
Looks maintainable, and I recognize it as a starting point, however, I would like to see explicit instructions to our team on how to add their status codes in comment form.

I am assuming from our conversations you want binary counting up, using the 8 bits counting left from the LSB, in the status code enum, keeping the subsystem flag intact.

But that shouldn't be left up to interpretation for the other people maintaining it.

So just add that and we are good.

Jebbs:
New changes since this was last worked on:
* Improved documentation which provides more details about the different parts of a status code and best practices on what to do when adding new codes
* Default cases for all switch statements which will let Mission Control know that it received an unknown status code or a code for an unknown sub system (this should catch anyone who forgot to add an associated message)

collinsc (approved):
Awesome thanks for adding the documentation

PR #32, opened 2018-02-09
Fiducial odom integration

collinsc:
# Description
We identified that there was a need for better testing facilities for sensor fusion and navigation. The existing fake odometry we could feed our libraries was not steady enough to make our mapping software or kahlman filters work in a predictable manner. 

So we identified using fiducial marker detection might be an option to allow us to refine our higher level algorithms.

# Statement of Work
This involved a couple of tweaks:
- _Format output of aruco action server_: I converted the output of the aruco action server from the opencv standard to ros standard coordiante systems. To allow us to use the measured data.
- _Integrate Aruco with existing tf2 coordinate system_: We use tf2 to keep track of our coordinate frames. To fit with the tf2 one parent rule. I decided making the bin the child of `odom` and parent of `base_footprint` was the most appropriate.
- _Fake some odometry_: The next step was to do some linear algebra to fudge some odometry. This was pretty easy to do with the built in utilities in tf2.
- _Create a testing wrapper_: So after we did all this I needed to write some logic that would localize the rover on startup and publish some odometry data. Which it does. This is contained in the `fiducial_odom_test.launch` in `tfr_launch` package.

# Results

So the results are pretty good from about 0.25 -0.75 meters with our existing aruco sheet. Which should be enough to get a deeper dive in testing navigation and sensor fusion. 

Here is a snapshot of our coordinate frames:
screenshot from 2018-02-09 09-20-03

Here is a snapshot of the captured data:
screenshot from 2018-02-07 23-23-41

# TODO
So I basically did all the math to actually go from aruco server output to localized bin. So on a seperate pull request we should reuse that logic in the fiducial action server. The logic is in the `bin_localizer.cpp` in  the localize function. 

# Testing
This is all test code, and very non portable, so make sure I didn't make any glaring oversights, but you won't be able to run or unit test this in a way that's very beneficial.


src/tfr_sensor/src/fiducial_odom_publisher.cpp
```diff
@@ -0,0 +1,277 @@
...
+        {
+            image_transport::ImageTransport it{n};
+            subscriber = it.subscribeCamera("image",10, 
+                    &FiducialOdom::process_odometry, this);
+            publisher = n.advertise<nav_msgs::Odometry>("odom", 10 );
+            ROS_INFO("Fiducial Odom Publisher Connecting to Server");
+            client.waitForServer();
+            ROS_INFO("Fiducial Odom Publisher Connected to Server");
+            //fill transform buffer
+            ros::Duration(2).sleep();
```
AdamuKaapan, line 67: Just making sure, a delay this long won't have much if any negative implications for any other systems depending on this?

src/tfr_sensor/src/fiducial_odom_publisher.cpp
```diff
@@ -0,0 +1,277 @@
...
+                odom.twist.twist.angular = tf2::toMsg(rpy_deltas/delta_t);
+
+                odom.twist.covariance = {  5e-3,   0,   0,   0,   0,   0,
+                                              0,5e-3,   0,   0,   0,   0,
+                                              0,   0,5e-3,   0,   0,   0,
+                                              0,   0,   0,5e-3,   0,   0,
+                                              0,   0,   0,   0,5e-3,   0,
+                                              0,   0,   0,   0,   0,5e-3};
+
+                //fire it off! and cleanup
```
AdamuKaapan: Love the

src/tfr_aruco/src/aruco_action_server.cpp
```diff
@@ -97,33 +98,37 @@ class TFR_Aruco {
...
+             *  also the coordinate axist for the aruco are in a different
+             *  coordinate system and are rotated here.
+             * */
+            result.relative_pose.pose.position.x = boardTransVec[2];
+            result.relative_pose.pose.position.y = boardTransVec[0];
+            result.relative_pose.pose.position.z = boardTransVec[1];
+            //let tf do the euler angle -> quaternion math
+            tf2::Quaternion rotated{};
+            //change rotated perspective RPY aruco output to ros coordinate system (2d)
+            rotated.setRPY(boardRotVec[2], boardRotVec[0], 3.1415 + boardRotVec[1]);
```
AdamuKaapan: Isn't there a pi constant somewhere? Not a huge issue but might be nicer to use that

AdamuKaapan (dismissed):
Everything looks good to me, I left one comment asking a question just to be sure and another comment on such a tiny issue it's really not worth worrying about much.

src/tfr_sensor/src/fiducial_odom_publisher.cpp
collinsc: This only impacts our system boot time, and we do it to suppress warning error messages, during tf boot which is normal. Since it is sleeping, that means the rest of our boot sequence can execute unhindered. So it shouldn't effect us too much.

src/tfr_aruco/src/aruco_action_server.cpp
collinsc: I can do that as a variable. Surprisingly a constant for pi is not in the `c++` standard, and ros doesn't have one either.

collinsc: There is one in the microsoft version of `cmath`, but it's not portable.

collinsc:
@AdamuKaapan I made your changes and fixed the whitespace in `aruco_action_server`

src/tfr_sensor/src/fiducial_odom_publisher.cpp
```diff
@@ -0,0 +1,277 @@
...
+                odom.child_frame_id = footprint_frame;
+                //get our pose and fudge some covariances
+                odom.pose.pose = relative_pose.pose;
+                odom.pose.covariance = {  5e-3,   0,   0,   0,   0,   0,
+                                             0,5e-3,   0,   0,   0,   0,
+                                             0,   0,5e-3,   0,   0,   0,
+                                             0,   0,   0,5e-3,   0,   0,
+                                             0,   0,   0,   0,5e-3,   0,
+                                             0,   0,   0,   0,   0,5e-3};
+
```
RyanBerge, line 186: Is this covariance matrix made up?  Or do these numbers come from somewhere?

RyanBerge (approved):
I did have one question, but I don't think the answer changes anything.

I can't really test this without markers and a camera that works, but I trust that's been done plenty.  I didn't notice any obvious programming errors.

src/tfr_sensor/src/fiducial_odom_publisher.cpp
collinsc: It's made up. In fact all of our covariance matrices are made up from completely arbitrary numbers (none of our device drivers shipped with them, I had to add them manually). The place where they matter is in `robot_localization`, it uses them as part of it's state estimation algorithm. Populating these with semi reasonable values is a `//TODO` for our sensor fusion team, as part of tuning the kahlman filter.

PR #33, opened 2018-02-11
Arm Action Server

AdamuKaapan:
# Description
Adds the Arm Action Server, which serves the purpose of taking in target joint angles for the arm, interfacing with the `move_group` node provided by MoveIt!, and then returning the result of the execution. Also provides a test client (which will be described later).

# TODO
The one remaining part of this component is determining how the move_group "move()" method returns errors. In this case, this would mean that a plan was successfully created, but the motors themselves failed in executing the motion (for example, the scoop got stuck in the ground). I have yet to find any documentation on this, and a quick test produced no results: it could be an exception, return value, or some other method that I don't know about.

# Testing
To test this there are two launch files that must be used in tandem. First, start `roslaunch tfr_control test_armserver.launch`. This starts all of the MoveIt! nodes, as well as the ArmActionServer and Rviz to view the robot state. Second, start `roslaunch tfr_control test_armclient.launch`. This should open a logger window to view ROS console output, as roslaunch for some reason doesn't show logger output in the console and it makes me sad.

## Expected Results
1. You should see the robot appear in Rviz upon running the first launch file.
2. Once you open the second launch file, you should see the arm begin to move in an approximated digging motion, and the console window should begin to show results.
3. After the arm dumps into the bin, it will attempt to make an invalid move (where it would clip into the bin), upon which there should be an error in the console window and the arm should stay where it is.
4. The test client should exit.


src/tfr_control/CMakeLists.txt
```diff
@@ -43,6 +44,7 @@ include_directories(
 add_executable(proto_driver_station 
   src/proto_driver_station.cpp
 )
+add_dependencies(proto_driver_station tfr_msgs_gencpp)
```
collinsc, line 47: Not a required change, but to satisfy my curiosity, what messages does this generate and why?

src/tfr_control/config/test_client_positions.yaml
```diff
@@ -0,0 +1,10 @@
+# Holds all of the positions for the test client to travel to (order doesn't
+# matter in this file, but this is the correct order for a dig cycle)
```
collinsc: This file is fine, but I have a naming question about the folder it's stored in. 

We never established any convention or standardization for this, but this folder is called config.

Here is the naming convention I have been using:
- launch
  - launch files
- src
  - c++/python code
- include
  - headers
- params
  - launch parameters for really long and complicated nodes
- test
  - google test files

we never really picked a standard for this, but we should talk about it monday.

Maybe for files like this, a data folder?
I don't know, let me know what makes sense to you.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
+/****************************************************************************************
+ * File:    arm_action_server.cpp
+ * Node:    arm_action_server
+ * 
+ * Purpose: This ActionServer is what handles the movement of the arm to given points.
+ *          It's tasked with interfacing with the move_group node to execute these
+ *          actions.
+ * 
+ *          This file includes <tfr_msgs/Arm.h>, which is one of seven headers
+ *          built by catkin from `tfr_msgs/action/Example.action`:
```
collinsc: Is this comment talking about the arm action or example action, it's a little unclear.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
+/****************************************************************************************
+ * File:    arm_action_server.cpp
+ * Node:    arm_action_server
+ * 
+ * Purpose: This ActionServer is what handles the movement of the arm to given points.
+ *          It's tasked with interfacing with the move_group node to execute these
+ *          actions.
+ * 
+ *          This file includes <tfr_msgs/Arm.h>, which is one of seven headers
```
collinsc: Also you do this below
`#include <tfr_msgs/ArmMoveAction.h>`

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
...
+ *              devel/include/tfr_msgs/ExampleGoal.h
+ *              devel/include/tfr_msgs/ExampleResult.h
+ * 
+ ***************************************************************************************/
+
+#include <actionlib/server/simple_action_server.h>
+#include <tfr_msgs/ArmMoveAction.h>
+#include <moveit/move_group_interface/move_group_interface.h>
+
+typedef actionlib::SimpleActionServer<tfr_msgs::ArmMoveAction> Server;
```
collinsc: It's not wrong, but in recent years, the `using` keyword, has started coming more into favor. It provides better template interoperability, and interoperability with some language constructs. 

However it basically accomplishes the same task of `typedef`, it's not a required change, I'll leave it up to you. 

Here is a little light reading for you:
https://stackoverflow.com/questions/10747810/what-is-the-difference-between-typedef-and-using-in-c11

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
...
+    ros::NodeHandle n;
+
+    // An async spinner is required here for the MoveIt setup to connect properly
+    ros::AsyncSpinner spinner(1);
+    spinner.start();
+
+    // Initialize the move_group connection (tl;dr sets up a ton of action server
+    // connections internally)
+    moveit::planning_interface::MoveGroupInterface mg("arm_end");
+    move_group = &mg;
```
collinsc: So you store a raw pointer as a global variable, that is used in an action server with an execute callback. 

Execute callbacks get started threaded by default so you are entering a potential shared memory "danger zone", you might not have made any mistakes, but the people maintaining your code might not be as dilligent. Global variables and threading don't play nicely. 

Probably what would make sense is to have a helper object or class that can hold
```
moveit::planning_interface::MoveGroupInterface *move_group;
const robot_state::JointModelGroup *joint_model_group;
```
as `private` member variables.


Here is how I would write it, but you do you. My one requirement is avoid the unscoped publicly accessible global variables in a threaded environment.

```
class MyClass
{
public:
  MyClass(ros::NodeHandle &n) :
      move_group{"arm_end"},
      joint_model_group{move_group.getCurrentState()->getJointModelGroup("arm_end")},
      server{n, "move_arm", boost::bind(&MyClass::execute, this, _1), false}
  { server.start(); }
  
...
private:
...
void execute(const tfr_msgs::ArmMoveGoalConstPtr& goal)
{
...
server.setSucceeded(foo);
}
...
  moveit::planning_interface::MoveGroupInterface &move_group;
  const robot_state::JointModelGroup &joint_model_group;
  actionlib::SimpleActionServer<tfr_msgs::ArmMoveAction>  &server;
}
}

int main(...)
{
...
ros::NodeHandle n;
ros::AsyncSpinner spinner(1);
spinner.start();
...
MyObj obj(n);
while (ros::ok()) ;
return 0;
}
```

src/tfr_control/src/test_client.cpp
```diff
@@ -0,0 +1,97 @@
...
+    
+    // Set up the action server and connect
+    Client client("move_arm", true);
+
+    ROS_INFO("Connecting to action server...");
+    client.waitForServer();
+    ROS_INFO("Connection established with server.");
+
+    // Loop through all of the positions to travel to and then exit
+    for (auto iter = position_names.begin(); iter != position_names.end(); iter++) {
```
collinsc: This is a perfect situation for a `ranged for` or `foreach`:
- http://en.cppreference.com/w/cpp/language/range-for
- http://www.cplusplus.com/reference/algorithm/for_each/

But "do something for every element in a collection" is a place this stuff can cut down of the boilerplate.

However this is test code so I don't really care and the change is optional.

src/tfr_control/src/test_client.cpp
```diff
@@ -0,0 +1,97 @@
...
+        // Get the name of the parameter in the positions namespace
+        std::string param_name = "positions/" + *iter;
+        ROS_INFO("Loading value: %s", param_name.c_str());
+
+        // Load the parameters (an array of doubles which represent joint angles)
+        std::vector<double> angles;
+        if (!n.getParam(param_name, angles)) {
```
collinsc: In the future if these are parameters for just that node, we should prefer private namespace scope, to avoid namespace pollution.

collinsc: http://wiki.ros.org/roscpp/Overview/Parameter%20Server
Read about how to access and use private parameters.

collinsc: However again, test code so not a required change

src/tfr_control/src/test_client.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // The list of positions to travel to, in order
+    std::vector<std::string> position_names = { "ready", "dig", "scoop", "out", "dump", "release", "invalid" };
+
+    // Basic ROS setup
+    ros::init(argc, argv, "example_client");
+    ros::NodeHandle n;
+
+    // Needed to spin while we're waiting for a blocking call later (waiting for
+    // server finished response)
+    ros::AsyncSpinner spin(1);
```
collinsc: Does everything that interacts with the moveit library need to be spun asynchronously?

collinsc (changes requested):
I asked a few questions, made a bunch of optional suggestions, and had one required change. There is a section where you need to remove some future threading risks.

Other than that looks good! It was really fun to get to watch it move!

I found some strange logs and errors, but those are in another comment.

collinsc:
@AdamuKaapan Side Note: besides the expected warnings described in your testing report. I get some extra warining messages.
roslaunch tfr_control test_armclient.launch
image

example:
```
Dropping first 1 trajectory point(s) out of 33, as they occur before the current time.
First valid point will be reached in 0.545s.
```
Pull up rqt console, and view all transmitted messages, while running. Let me know if you see these. It it's just me whatever that's fine.

also this message was displayed to `stdout` when I ran `roslaunch tfr_control test_armserver.launch`
```
[ERROR] [1518427054.745425142]: Got a callback on a goalHandle that we're not tracking.                 This is an internal SimpleActionClient/ActionClient bug.                 This could also be a GoalID collision

```

src/tfr_control/CMakeLists.txt
AdamuKaapan: This currently uses messages from a long time ago, so I'd be fine with removing the whole executable itself. I just went through and added this to all the executable builds to ensure that they don't build without messages in case they need them (which most of them usually will because, hey, we're working in ROS).

src/tfr_control/config/test_client_positions.yaml
AdamuKaapan: Yeah, I agree, this would be a good file to move. `data` seems like a good name for this case, I agree

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: Oh, that was a copy-pasta error, I'll be sure to fix that

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: Thanks, I'll fix that as well

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: I mainly was just copying from the sample action server in tfr_util. For now I'll stick with what I have for consistency (although I'm completely fine with either, it makes no difference).

src/tfr_control/src/test_client.cpp
AdamuKaapan: Good note, I'll do that for the future. Might do it now just for a test/practice.

src/tfr_control/src/test_client.cpp
AdamuKaapan: I'll give it a try just to get familiar with it.

AdamuKaapan:
@collinsc I see those errors as well, I believe that's an error with MoveIt or ros_control and not something we can control. I'll do some research to be sure.

AdamuKaapan:
@collinsc Although that second error I'm not as sure about, I'll investigate that one more

src/tfr_control/src/test_client.cpp
AdamuKaapan: I don't think everything does (might need some more testing on that), but it seems that instantiating MoveIt needs to be.

src/tfr_control/CMakeLists.txt
collinsc: Well all of the message generation dependenices are in the `tfr_msgs` package. To use the messages you don't need to take a dependency on message generation, only `tfr_msgs`

src/tfr_control/src/arm_action_server.cpp
collinsc: Yup it's a small thing. I ended up not using the typdef at all in my code, to each their own.

src/tfr_control/src/test_client.cpp
collinsc: `for (auto iter = position_names.begin(); iter != position_names.end(); iter++)`
can become:
`for (auto &position : position_names)
{
....
}`

src/tfr_control/CMakeLists.txt
AdamuKaapan: Hmmm, interesting, I'll need to check if that works to be sure, but thanks for pointing that out.

src/tfr_control/src/test_client.cpp
AdamuKaapan: Oooh, that's perfect. I'll use that.

src/tfr_control/CMakeLists.txt
RyanBerge: @collinsc Supposedly this was a fix to a build problem Adam was having a few weeks ago, but that problem never quite made sense to me.

src/tfr_control/src/arm_action_server.cpp
RyanBerge: I go back and forth.  `using` is a newer standard and is probably preferred, but either way is fine.

src/tfr_control/config/test_client_positions.yaml
RyanBerge: `config` and `data` both seem fine to me.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
...
+    joint_group_positions[1] = goal->pose[1];
+    joint_group_positions[2] = goal->pose[2];
+    joint_group_positions[3] = goal->pose[3];
+
+    // Set the current target
+    move_group->setJointValueTarget(joint_group_positions);
+    moveit::planning_interface::MoveGroupInterface::Plan my_plan;
+
+    // Try to plan to the given target
+    bool success = (move_group->plan(my_plan) == moveit::planning_interface::MoveItErrorCode::SUCCESS);
```
RyanBerge: This is definitely optional, but when you have something as long as `moveit::planning_interface::MoveItErrorCode::SUCCESS` I think adding a `typedef` or `using` statement is perfectly reasonable.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    if (success) {
+        // Planning was successful, actually execute the movement
+        ROS_DEBUG("Executing movement");
+        move_group->move();
+        // TODO: How are errors handled?
+        //  - Further note: in the case that, say, the arm gets stuck and can't
+        //    move, how are these errors returned? Exceptions, return code, or
+        //    something else?
```
RyanBerge: You have a comment that asks how errors are handled.  Are you asking how the `move` function handles errors, or how we want to handle them?

src/tfr_control/src/arm_action_server.cpp
RyanBerge: I don't think it has to be a class, but I agree that it shouldn't be stored globally here with no shared-data protections.  I don't think there's a race condition in the code you've written, but that doesn't mean it couldn't be introduced later by accident.

RyanBerge (changes requested):
I agree with pretty much everything @collinsc said, and I asked a couple of questions of my own.

src/tfr_control/config/test_client_positions.yaml
collinsc: I think they both work also, but let's pick a naming convention eventually.

src/tfr_control/src/arm_action_server.cpp
collinsc: Functionally the same, however `using` can help you avoid one particularly nasty template use case. See above article.

src/tfr_control/src/arm_action_server.cpp
collinsc: What alternatives to a class would you suggest?

src/tfr_control/CMakeLists.txt
collinsc: Weird, that is bizarre. also this has been bugging me. @AdamuKaapan the `proto` prefix was something we used waaaaayyyyy back when we were porting stuff over from the prototyping code, and went out of use very shortly after the start of development. Are we using `proto_driver_station` for anything, or is that a piece of test code, or something from a previous sprint?

AdamuKaapan: `proto_driver_station` was something you added when you did the initial gazebo commit, yeah. We don't need this executable any more, we can remove it in another PR soon.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: Thanks, that's a good idea: went ahead and implemented it in an upcoming commit.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: I'm asking how the move function handles errors: I know how I want to handle them I think, it'll just be returning a failure and maybe stopping execution of the arm command.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: I think I'm going to go with a class, @collinsc showed me that pattern earlier and it does seem really clean and helps solve some other issues I have sometimes.

src/tfr_control/src/test_client.cpp
AdamuKaapan: Implemented this in the upcoming commit.

src/tfr_control/src/arm_action_server.cpp
RyanBerge: @collinsc You can also create static variables inside of free functions, which are basically scoped globals, but allow for access guards.  I don't really think that's better, but it does work.

collinsc: For the sake of my own curiosity, could you put that in pseudocode?

src/tfr_control/CMakeLists.txt
collinsc: Thanks, I'll make a trello card for that.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: ... took me literally 30 seconds to find the answer to this, so I'm good now. Added a quick super simple implementation of it, incoming in like 5 minutes.

collinsc (dismissed):
Looks great! thanks for implementing that.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -6,77 +6,93 @@
...
+            // Planning was successful, actually execute the movement
+            ROS_DEBUG("Executing movement");
+            moveSuccess = (move_group.move() == MoveItErrorCode::SUCCESS);
+
+            if (!success) {
+                ROS_WARN("Planning sucessful, but execution failed");
+            }
+        } else {
+            ROS_WARN("Planning of movement failed, not executing");
+        }
```
RyanBerge: Current coding standard says braces get their own line.

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -6,77 +6,93 @@
...
+        // Send the result message and set the appropriate action server status
+        // if both planning and execution was successful. It may be worthwile to
+        // add a distinction here later (aka "did the motors fail? or is this
+        // literally somewhere we aren't allowed to move?"), but that's for
+        // later if we determine we need it.
+        if (success && moveSuccess) {
+            server.setSucceeded(result);
+        } else {
+            server.setAborted(result);
+        }
```
RyanBerge: See above.

RyanBerge (changes requested):
Refactored class is much better.  Just a couple nitpicks about style.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: Thanks. I'm pretty sure I did this on accident in a few other places, so I'll quickly scan all of my files to be sure (already caught one in my constructor as well).

collinsc (approved):
redoing stale review

PR #34, opened 2018-02-14
Dumping aruco integration

collinsc:
# Prerequisite
#32 

# Description
This pull request integrates aruco with the dumping action server, it's pretty short.

Also I did a small readability refactor. Since it didn't really make sense to include the dumping action server anywhere, and it's one and only job is to interact it's easiest for me to read and understand what is going on it it's all together in one file. 

Basically class definition at the top of a file, and main loop using it at the bottom. This is common to a lot of simple ros code out there. 

# Testing
Since this isn't portable, I figured I would try something new and do a video demo. 
https://youtu.be/zMWfCg-NI5A


RyanBerge (dismissed):

RyanBerge (approved):

Jebbs (approved):
Looks good to me. I also just noticed that you end your multi-line comments with `* */`. Weird.

PR #35, opened 2018-02-15
Sensor fusion

collinsc:
# Prereqs
#32 
# Description
This is the base setup for the `robot_localization` library. It takes in data from the:
- __xsens imu__:
  - orientation
  - linear velocity
  - linear accerlation
  - angular acceleration
- __duo3d imu__:
  - orientation
  - linear velocity
  - angular acceleration
- __fiducial odometry__:
  - position
  - orientation

It fuses these data streams into the standard output of position for robot localization on the `/odometry/filetered` topic. It also publishes transforms to `tf`.

This is the base completely vanilla not optimized implementation of the library with 2d settings on. It represents a couple of work sessions with @moyaog  and @aabdin02. We decided to do a check in here because it is stable enough for other branches to test on now.

# Testing
None of us have done quantitative testing of the algorithm, but we have done some qualitative testing. 
I did a video demo for the review team's benefit. 
https://youtu.be/K4qaPaj9OPg


RyanBerge (dismissed):
I'm not qualified to speak to the correctness of the configuration files, which is most of this PR, but the code that is in here looks fine to me.

RyanBerge (dismissed):
(re-approving after stale review was dismissed)

RyanBerge (approved):

moyaog (approved):

PR #39, opened 2018-02-24
Gui

collinsc:
# Description
The GUI/Utility methods for publishing status codes to the `com` channel. It provides a qt framework application that resides in the `tfr_mission_control` package. It also provides a utility class in `tfr_utilities` called `status_publisher` which publishes to the control center and the desired ros visibility level simultaneously.

So let's talk about the features of the gui, it starts in an unitialized state. At this point the user can start the mission autonomously, or start it in teleop, which makes the system ready to accept commands and initializes the clock. 

The teleop button is hidden in the top corner to make it harder to access, because that should only be used for testing.

Running autonomy will just flash a bunch of messages on the screen simulating the system running. You can switch to teleop mode by stopping autonomy. Switching in between teleop and autonomy will print some error messaages to stdout. This is caused by me calling the `isActive` methods on the action clients when they don't have a goal, to synchronize the servers. This error message is erroneous, and is a perfectly valid use case for synchronizing in between the two action clients. I have tested this thoroughly, and the forum community seems to agree this is not really an error, and sometimes a usefull thing to do.

So on to teleop. In telop mode we want to process user commands. In general some actions like digging and dumping we want to "fire and forget", but some like turntable movement, and directional control, we want to do until the user stops holding the button, then stop doing it.

This screen also accepts `WASD` control. Because of some details I go into in the documentation for the code I need to do manual debouncing of keyboard events. This means the controls might throw in an erroneous stop message if your keyboard repeated key OS delay is set longer than 0.3 seconds. So for a smooth experience change your repeated keys delay to something smaller in the keyboard settings menu. As shown here, you can look it up with unity dash.
screenshot from 2018-02-23 18-39-17

To test the scrolling window try the `toggle motors` button. It is the only thing hooked up there right now. Just click it a bunch of times.
# Testing
Go ahead and fire up the GUI, and try and break it, let me know of anything weird!
`roslaunch tfr_launch mission_control_test`


src/tfr_executive/src/teleop_action_server.cpp
```diff
@@ -141,12 +159,21 @@ class TeleopExecutive
...
+            else if (code == tfr_utilities::TeleopCode::RESET_STARTING)
+            {
+                ROS_INFO("Teleop Action Server: Command Revieved, RESET_STARTING");
+                //all zeros by default
+                drivebase.publish(move_cmd);
+                //TODO hook up reset when ready
+            }else
             {
                 ROS_WARN("Teleop Action Server: UNRECOGNIZED COMMAND");
             }
```
RyanBerge: Is there a reason this function isn't built on a switch statement?

src/tfr_executive/src/teleop_action_server.cpp
```diff
@@ -97,13 +97,19 @@ class TeleopExecutive
         {
             geometry_msgs::Twist move_cmd{};
             auto code = static_cast<tfr_utilities::TeleopCode>(goal->code);
-            if (code == tfr_utilities::TeleopCode::NONE)
+            if (code == tfr_utilities::TeleopCode::STOP_DRIVEBASE)
             {
-                ROS_INFO("Teleop Action Server: Command Revieved, RESET");
+                ROS_INFO("Teleop Action Server: Command Revieved, STOP_DRIVEBASE");
```
RyanBerge: "Command Revieved"

src/tfr_executive/src/teleop_action_server.cpp
```diff
@@ -141,12 +159,21 @@ class TeleopExecutive
...
-            else
+            else if (code == tfr_utilities::TeleopCode::RESET_STARTING)
+            {
+                ROS_INFO("Teleop Action Server: Command Revieved, RESET_STARTING");
+                //all zeros by default
+                drivebase.publish(move_cmd);
+                //TODO hook up reset when ready
+            }else
             {
                 ROS_WARN("Teleop Action Server: UNRECOGNIZED COMMAND");
```
RyanBerge: We return a `succeeded` state on unrecognized commands, even though we take no action.  Should probably set `failure` here and return.

RyanBerge (changes requested):
Just a couple questions and comments.  One more, which I can't put in an inline comment because the area isn't part of the diff for this PR, is that the `frequency` value in `teleop_action_server` is not currently being used.  You should do one of the following:

- Remove it
- Add preemption checks
- Add TODO comments where preemption checks should be

I didn't look through the QT code very closely because I'm not familiar with the API. If there's something seriously wrong with it, we should notice during testing.

src/tfr_executive/src/teleop_action_server.cpp
collinsc: fixed, but we will be removing these on a future release, so they don't need to be too pretty

src/tfr_executive/src/teleop_action_server.cpp
collinsc: It was a while ago, but I think that was a relic of a typing logic error with enum classes. I put it in an else if telling my self "put it in a switch if you ever figure that out" then I figured it out and forgot to the logic it in the switch.

src/tfr_executive/src/teleop_action_server.cpp
collinsc: The state we want to signal that behavior is called `aborted` but I know what you mean, sure

collinsc:
@RyanBerge we will use that frequency value during digging/dumping, and that is a task for systems integration. 

Looking through my comments, it looks like I put a TODO there but it wasn't a very specific comment, I made it longer and made sure to mention that was the place we need to check for preemption

src/tfr_executive/src/teleop_action_server.cpp
```diff
@@ -97,58 +97,115 @@ class TeleopExecutive
...
+                        drivebase.publish(move_cmd);
+                        break;
+                    }
+
+                case (tfr_utilities::TeleopCode::DIG):
+                    {
+                        // TODO integrate and check for preemption
+                        ROS_INFO("Teleop Action Server: Command Recieved, DIG");
+                        tfr_msgs::DurationSrv digging_time;
+                        ros::service::call("digging_time", digging_time);
```
AdamuKaapan, line 171: Are we operating the digging as single digging sets (scoops) or just saying "keep digging until I tell you to stop"?

src/tfr_mission_control/plugin.xml
```diff
@@ -0,0 +1,19 @@
+<library path="lib/libtfr_mission_control">
+  <class name="tfr_mission_control/MissionControl"
+         type="tfr_mission_control::MissionControl"
+         base_class_type="rqt_gui_cpp::Plugin">
+    <description>
+      A GUI plugin to ...
+    </description>
+    <qtgui>
+      <group>
+        <label>New Group</label>
```
AdamuKaapan, line 10: We should probably fill these out: even though we probably won't ever see them, just for good practice (although no big deal if it doesn't get changed, so don't let this block the PR).

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
```diff
@@ -0,0 +1,409 @@
...
+        widget = new QWidget();
+        ui.setupUi(widget);
+        if (context.serialNumber() > 1)
+        {
+            widget->setWindowTitle(widget->windowTitle() +
+                    " (" + QString::number(context.serialNumber()) + ")");
+        }
+
+        //sets our window active, and makes sure we handle keyboard events
+        widget->setFocusPolicy(Qt::StrongFocus);
```
AdamuKaapan, line 62: I'm assuming that "Strong Focus" means that keyboard events outside the window don't trigger events here, correct?

AdamuKaapan: If that's the case that's a really good thing

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
```diff
@@ -0,0 +1,409 @@
...
+        ui.dig_button->setEnabled(value);
+        teleopEnabled = value;
+    }
+
+    /*
+     * Toggles the autonomy stop button
+     * */
+    void MissionControl::setAutonomy(bool value)
+    {
+        ui.teleop_button->setEnabled(value);
```
AdamuKaapan, line 216: The naming here is really confusing, just because I'm seeing "set autonomy" maps to "enable the teleop button", which feels really backwards to me. It just worries me that something's gonna get flipped, so I'd like to see a bit more documentation/a name change to make this clearer.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
```diff
@@ -0,0 +1,409 @@
...
+     * This triggers our custom emitStatus signal, which triggers the built in
+     * text update in the text box, a seperate timer event is constantly
+     * scrolling the window on a set fast interval.
+     * */
+    void MissionControl::updateStatus(const tfr_msgs::SystemStatusConstPtr &status)
+    {
+        auto str = getStatusMessage(static_cast<StatusCode>(status->status_code), status->data);
+        //need to wrap in signal safe "Q" object
+        QString msg = QString::fromStdString(str);
+        emit emitStatus(msg);
```
AdamuKaapan, line 317: Just 'cause I'm curious, is this some sort of special Qt preprocessor thing?

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+    }
+
+    //triggers state change into from teleop into autonomy
+    void MissionControl::goTeleopMode()
+    {
+        setAutonomy(false);
+        while (!autonomy.getState().isDone()) autonomy.cancelAllGoals();
+        setTeleop(true);
+        softwareStop();
```
AdamuKaapan, line 366: What's the reasoning for having softwareStop() so late in this method (vs. goAutonomousMode() where it's the first line)? I'm guessing it's for letting autonomy hand over control of the actuators first before we stop them?

AdamuKaapan (changes requested):
Nothing really breaking here, just had a few questions/little comments that I think should at least be considered before I actually approve it.

src/tfr_executive/src/teleop_action_server.cpp
collinsc: I give it the time allowed to dig just like the specification

src/tfr_mission_control/plugin.xml
collinsc: sure I don't pretend to be a gui expert, in fact quite the opposite. I just wanted something that turned on.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
collinsc: I also don't know I just followed the tutorial to get it to turn on, not a front end dev.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
collinsc: it's a window state, I think it makes sense. The only button you should be able to press when you are in autonomous mode is the enable teleop.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
collinsc: It's sends the string to the next event in the chain of events to process

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
collinsc: It's just an extra measure. Notice how we are letting the autonomous action server preempt itself? on line 364. I just did it there in case anything failed with the preemption. It doesn't hurt us.

collinsc:
@AdamuKaapan I answered your questsions

src/tfr_executive/src/teleop_action_server.cpp
RyanBerge: @AdamuKaapan Executive should just know how to tell it to go with a timer. The Mining System is in charge of deciding how many scoops that is. This looks correct to me.

RyanBerge (approved):

src/tfr_executive/src/teleop_action_server.cpp
AdamuKaapan: Yeah, that makes sense; I just wasn't sure if we wanted to do it a bit more granularly because it's manual operation, but I'm totally fine with this too.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
AdamuKaapan: Yeah, I was just wondering about the `emit emitStatus(msg);`, that seems like preprocessor stuff unless there's some weird `emit` keyword in C++ I don't know about, hahaha.

src/tfr_mission_control/src/tfr_mission_control/mission_control.cpp
AdamuKaapan: Makes sense, yeah, if everything else also stops the motor.

AdamuKaapan (approved):

PR #44, opened 2018-03-20
Localization Action Server

collinsc:
# Description
The localization action server, rotates the robot until it sees the marker, sets up the bin publisher. etc
There is a full specification of the code in `tfr_localization` `localization_action_server`
# Testing
It's not portable here are some screenshots.

When you run it this is what you see:
screenshot from 2018-03-19 17-08-39
You see in the right hand frame the twist output on `cmd_vel` rotating the robot in a circle, then stops.


Jebbs (dismissed):
No problems spotted. Looks like you'll have to rebase before this get's merged though.

Edit: You beat me to the rebase!

RyanBerge (dismissed):

PR #46, opened 2018-03-21
Adc driver

collinsc:
# Description
the arduino driver for the ADC. super short

video demo
https://youtu.be/v6srLKs1EBk


src/tfr_control/arduino/encoder/encoder.ino
```diff
@@ -16,26 +19,30 @@ const int GEARBOX_LEFT_B = 3;
...
 //encoders
 Quadrature gearbox_left(GEARBOX_CPR, GEARBOX_LEFT_A, GEARBOX_LEFT_B);
-tfr_msgs::EncoderReading reading;
+tfr_msgs::EncoderReading encoderReading;
 
-//algorithm constants
-const float rate(1/50.0);
-
-ros::Publisher encoders("encoders", &reading);
+ros::Publisher encoders("encoders", &encoderReading);
```
AdamuKaapan: Have you considered combining the encoders and pots into one single message? It might help to save on some weird header size overhead and reduce the total amount of data sent (although if you're planning on splitting it between arduinos then that doesn't matter).

AdamuKaapan (dismissed):
Looks good, just left a comment for you to consider which may or may not be relevant depending on what your plan is from here on out.

src/tfr_control/arduino/encoder/encoder.ino
collinsc: That's not a bad idea, it will actually make my code simpler.

collinsc: I'll add it in next iteration

RyanBerge (approved):

PR #48, opened 2018-03-21
Add setup files for networking support

Jebbs:
This commit adds two files in the top level directory which are intended to be the "boot" scripts for both the robot and mission control. These export the needed IP information needed for networking, and will later call the right roslaunch commands to initialize the systems.

I also split the mission control launch file into two parts as part of testing.


collinsc (approved):
easy peasy looks good

AdamuKaapan (approved):

RyanBerge (approved):

PR #49, opened 2018-03-23
Digging action server

AdamuKaapan:
# Description
This implements both the digging action server and the digging procedure queue. The digging queue is in charge of feeding digging "sets" (one set ~= one hole dig over four scoops) to the digging action server to process. The digging action server takes these sets, and sends them to the arm action server for processing. The digging action server, and therefore the arm action server, is entirely preemptable (AAS preemption is also included in this PR since it wasn't implemented before).

# Testing
A minor test has been tweaked from last time: `tfr_control test_armserver.launch` and `tfr_control test_armclient.launch`. The client now will preempt the server if you press any key in the terminal.

To test the digging action server, first run `tfr_mining test_diggingserver.launch` and let it all boot up until Rviz is shown. Then start `tfr_mining test_diggingclient.launch`. You should see the digging arm begin digging from the left, doing two dumps next to the robot and then two dumps into the bin. It should then proceed to do this for the next holes to the right. Once it's done digging, it should end in a straight upright position. If you press any key in the client terminal, it will preempt.


src/tfr_control/src/arm_action_server.cpp
```diff
@@ -52,17 +70,54 @@ class ArmActionServer {
...
+                    std::vector<double> current_pos;
+                    move_group.getCurrentState()->copyJointGroupPositions(&joint_model_group, current_pos);
+
+                    // Populate the result message
+                    tfr_msgs::ArmMoveResult result;
+                    result.pose.resize(4);
+                    result.pose[0] = current_pos[0];
+                    result.pose[1] = current_pos[1];
+                    result.pose[2] = current_pos[2];
+                    result.pose[3] = current_pos[3];
```
collinsc: what is this result used for?

src/tfr_mining/src/digging_set.cpp
```diff
@@ -0,0 +1,31 @@
+#include "digging_set.h"
+
+namespace tfr_mining
+{
+    DiggingSet::DiggingSet() : states{}, time_estimate{0}
+    {
+
+    }
```
collinsc, line 8: Weird, whitespace non mandatory change.

src/tfr_mining/src/digging_set.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+    void DiggingSet::insertState(std::vector<double> state, double time)
+    {
+        states.push(state);
+        time_estimate += time;
```
collinsc, line 13: non mandatory but the ros::Time object has some neat convenience methods for stuff like this.

src/tfr_mining/include/tfr_mining/digging_queue.h
```diff
@@ -0,0 +1,45 @@
...
+        bool isEmpty();
+
+        /**
+         * Returns the next set in the queue and removes it.
+         **/
+        DiggingSet popDiggingSet();
+    private:
+        std::queue<DiggingSet> sets;
+
+        void generateDigAndDump(ros::NodeHandle &nh, DiggingSet &set, double rotation, int dig_number);
```
collinsc: comment description of what this does in header please

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    void DiggingQueue::generateDigAndDump(ros::NodeHandle &nh, DiggingSet &set, double rotation, int dig_number)
+    {
+        // Ready (constant no matter what)
+        std::vector<double> ready_pos;
+        double ready_time;
+        if (!nh.getParam("positions/ready", ready_pos) || !nh.getParam("positions/ready_time", ready_time)) {
+            throw std::runtime_error("Error loading ready state");
+        }
+        std::vector<double>::iterator it = ready_pos.begin();
```
collinsc: the auto keyword can make iterators be a lot  less verbose, style suggestion

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    DiggingSet DiggingQueue::popDiggingSet()
+    {
+        DiggingSet set = sets.front();
+        sets.pop();
+        return set;
+    }
+
+    void DiggingQueue::generateDigAndDump(ros::NodeHandle &nh, DiggingSet &set, double rotation, int dig_number)
+    {
```
collinsc, line 37: This seems like important business logic and deserves a description comment

collinsc: some parameter description would be nice here too.

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+        ready_pos.insert(it, rotation);
+        set.insertState(ready_pos, ready_time);
+
+        // Dig position
+        std::vector<double> dig_pos;
+        double dig_time;
+        std::stringstream dig_str, dig_time_str;
+        dig_str << "positions/dig" << dig_number;
+        dig_time_str << dig_str.str() << "_time";
+        if (!nh.getParam(dig_str.str(), dig_pos) || !nh.getParam(dig_time_str.str(), dig_time)) {
```
collinsc: A quick note in this method about where these parameters are located, what the file is called and how it is loaded will be useful.

src/tfr_mining/src/test_digging_client.cpp
```diff
@@ -0,0 +1,107 @@
...
+ ***************************************************************************************/
+
+#include <tfr_msgs/DiggingAction.h> // Note: "Action" is appended
+#include <actionlib/client/simple_action_client.h>
+#include <termios.h>
+
+typedef actionlib::SimpleActionClient<tfr_msgs::DiggingAction> Client;
+
+// Non-blocking input function (useful for testing preemption)
+char getch()
```
collinsc, line 26: I have used this exact same code copypasta from stack overflow before, lol.

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+        //    to the left. We'll fill in our previous holes a bit as we go, and
+        //    it ensures that we never put digging material in a place we'll dig
+        //    later
+        for (int j = 0; j < 4; j++)
+        {
+            DiggingSet set;
+            generateDigAndDump(nh, set, 3.14159265 / 4 - ((3.14159265 / 4) * j), 1);
+            generateDigAndDump(nh, set, 3.14159265 / 4 - ((3.14159265 / 4) * j), 2);
+            generateDigAndDump(nh, set, 3.14159265 / 4 - ((3.14159265 / 4) * j), 3);
+            generateDigAndDump(nh, set, 3.14159265 / 4 - ((3.14159265 / 4) * j), 4);
```
collinsc: what are these magic numbers get passed in as parameters, a comment is needed here

src/tfr_mining/src/digging_action_server.cpp
```diff
@@ -0,0 +1,169 @@
...
+        ros::Time startTime = ros::Time::now();
+        ros::Time endTime = startTime + goal->diggingTime;
+        ROS_INFO("Starting at %d, ending at %d", startTime.sec, endTime.sec);
+
+        Client client("move_arm", true);
+        ROS_DEBUG("Waiting for arm action server...");
+        client.waitForServer();
+        ROS_DEBUG("Connected with arm action server");
+
+        // TODO: Handle preemption!
```
collinsc: I think you have finished this

src/tfr_mining/src/digging_action_server.cpp
```diff
@@ -0,0 +1,169 @@
...
+    {
+        server.start();
+        //server.registerPreemptCallback(boost::bind(&DiggingActionServer::preempted, this));
+    }
+
+private:
+    /*void preempted()
+    {
+        ROS_WARN("Digging preempted");
+    }*/
```
collinsc: Why do we have commented out methods in code we are checking in?

src/tfr_mining/src/digging_action_server.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+typedef actionlib::SimpleActionServer<tfr_msgs::DiggingAction> Server;
+typedef actionlib::SimpleActionClient<tfr_msgs::ArmMoveAction> Client;
+
+class DiggingActionServer {
+public:
+    DiggingActionServer(ros::NodeHandle nh, ros::NodeHandle nh_priv) : nh_private{nh_priv}, queue{nh_priv}, server{nh, "dig", boost::bind(&DiggingActionServer::execute, this, _1), false}
+    {
+        server.start();
+        //server.registerPreemptCallback(boost::bind(&DiggingActionServer::preempted, this));
```
collinsc: see below

src/tfr_control/src/arm_action_server.cpp
```diff
@@ -86,18 +141,35 @@ class ArmActionServer {
...
+        // getting the state and processing fast enough)
+        ros::Duration(0.5).sleep();
+
+        if (success && dig_status == 0)
         {
+            ROS_INFO("Successful!");
             server.setSucceeded(result);
         } else
         {
+            ROS_INFO("Unsuccessful!...");
```
RyanBerge: "Unsuccessful!..."

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+            double release_time;
+            if (!nh.getParam("positions/excessrelease", release_pos) || !nh.getParam("positions/excessrelease_time", release_time)) {
+                throw std::runtime_error("Error loading excessrelease state");
+            }
+            it = release_pos.begin();
+            release_pos.insert(it, rotation + (3.14159265 / 4));
+            set.insertState(release_pos, release_time);
+        }
+    }
+}
```
RyanBerge, line 145: Your braces are not up to the standard in this file.

collinsc (changes requested):
On the whole it looks good, the only changes that I think are mandatory are some places I think you need some clarifying documentation. Add those in and I think we are in shape. Good Job on this PR, I know it was a lot of work.

src/tfr_mining/src/digging_queue.cpp
```diff
@@ -0,0 +1,120 @@
...
+        return set;
+    }
+
+    void DiggingQueue::generateDigAndDump(ros::NodeHandle &nh, DiggingSet &set, double rotation, int dig_number)
+    {
+        // Ready (constant no matter what)
+        std::vector<double> ready_pos;
+        double ready_time;
+        if (!nh.getParam("positions/ready", ready_pos) || !nh.getParam("positions/ready_time", ready_time)) {
+            throw std::runtime_error("Error loading ready state");
```
RyanBerge: How are we dealing with exceptions? I don't remember talking a lot about error handling for things like this (mostly we've been using default parameters). Are you expecting this to crash the node?

RyanBerge (changes requested):

src/tfr_mining/src/digging_queue.cpp
collinsc: I have tried to avoid exceptions when I can. Usually I just print to ROS_WARN with configuration level stuff like this, it isn't a mission failing bug, it's just us being an idiot and mistyping something.

RyanBerge: I don't have a strong opinion here. This isn't something the robot can recover from without us going in and fixing something.

collinsc: Either way warns us that we are doing something stupid. Personally I would do the warning just to stay consistent with the rest of the code, but this version does the job so I don't think it's a big deal at all.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: It was in the spec, so I went ahead and implemented it. It's currently used for nothing as far as I know, so it could probably be removed.

src/tfr_mining/src/digging_set.cpp
AdamuKaapan: What's the issue you're seeing?

src/tfr_mining/src/digging_action_server.cpp
AdamuKaapan: Why indeed I have!

src/tfr_mining/src/digging_action_server.cpp
AdamuKaapan: Good catch, that's a remnant from when I was trying to work out how preemption worked before you showed me.

src/tfr_control/src/arm_action_server.cpp
AdamuKaapan: Fiiiiiine I'll change that

src/tfr_mining/src/digging_queue.cpp
AdamuKaapan: I don't believe I used exceptions anywhere else, so I'll just print a warning instead and return early as to prevent unexpected behavior

src/tfr_control/src/arm_action_server.cpp
collinsc: Yeah, I think that should be on the chopping block, There are definitely some parts of the spec that we wrote that didn't make sense in hindsight. However if you don't feel like fixing it now, it's not breaking anything, so you can throw it in the maintenence queue.

collinsc: As a trello card

src/tfr_mining/src/digging_set.cpp
collinsc: Usually with methods with an empty body I see people just do `{}` but that is just a style thing

collinsc (dismissed):

collinsc (approved):

RyanBerge (approved):

PR #50, opened 2018-03-23
Control Layer

collinsc:
# Prequisites
#46 
# Description
The hardware layer of the control stack is completely finished and ready to integrate. It's all manual testing and hardware so I will do this with screenshots.

I added in all the code to pack and unpack the information from our messages and send it to the controllers, then write commands back to the motors. There are a few features of note to see here.

_Position Controllers_ I set up all of our potentiometers to read in position, which is what you are supposed to do with potentiometers, and refactored our control layer to output position in the bin, turntable, and arm. 

_Bin Controller_ the bin controller was a big question mark, and it turned out being a one line fix we have full functionality

_Arduino Update_ I use a modified driver for the adc and set up a pipeline to allow us to process messages from both boards at once. Using this we can meet our target response time of 30 hz.

_Acceleration Limiter_ mechanical has given us the requirement of keeping acceleration to a 1 m/s^2 limit. I have acceleration limits in my application to avoid skidding out, but I added an extra safety at the control layer. This will help our teleop users.

_Potentiometer Units_ Our potentiometers now read position in radians to match the joints they will be measuring, with updated figures from mechanical.

# Tests
To test our acceleration limiter I generated commands oscilating the robot back and forth wildly. You can see the limited result on the right collumn.
limitation

To test our potentiometer units I set the lower limit for my test potentiometer to be 0.20944 rads, and the upper to be 1.693 rads. These are close to what we will actually see out of that joint. 

Observe the low reading
bin test

And the high
hihg


src/tfr_control/arduino/encoder/encoder.ino
```diff
@@ -1,41 +1,121 @@
...
+  {4.88,  0.000, 0.786}       //BIN_RIGHT
+};
+
+
 //encoders
-Quadrature gearbox_left(GEARBOX_CPR, GEARBOX_LEFT_A, GEARBOX_LEFT_B);
-tfr_msgs::EncoderReading reading;
+VelocityQuadrature gearbox_left(CPR, GEARBOX_LEFT_A, GEARBOX_LEFT_B);
+//TODO Quadrature gearbox_right(CPR, GEARBOX_LEFT_A, GEARBOX_LEFT_B);
+//TODO Quadrature turntable(CPR, GEARBOX_LEFT_A, GEARBOX_LEFT_B);
```
RyanBerge: This is just because we don't have things to plug in yet?

RyanBerge (dismissed):

src/tfr_control/arduino/encoder/encoder.ino
collinsc: Yup!

collinsc: In theory when we plug in the the hardware devices, and wire them we should be able to uncomment the line, recompile and be good to go. I have tested everything I can on a comprable device.

RyanBerge (dismissed):

src/tfr_control/arduino/encoder/encoder.ino
```diff
@@ -2,93 +2,120 @@
...
 {
-    const static float ADC_RANGE=0.0001875;
-    float extension_range; //the max range the potentiometer can be extended
-    float max_voltage; //the max voltage for the potentiometer
-    float getExtension(int16_t val)
-    {
-        return val* ADC_RANGE/max_voltage; 
-    }
+   const static float ADC_RANGE=0.0001875;
+   float max_voltage; //the max voltage for the potentiometer
```
AdamuKaapan, line 27: Does the mixing of floats and doubles have a reason? (could everything be a double? I know you mentioned some library calls need it) It bugs me stylistically a bit, but if there's reasons for it then it's fine.

src/tfr_control/arduino/quadrature.h
```diff
@@ -53,11 +53,37 @@ class Quadrature
...
 
     EncoderData data;
 
     Encoder encoder;
     
 };
+/*
+  Class for efficiently reading quadrature encoder data, instantiate, and
+  call getVelocity in the main loop. This is designed to be used in a system 
+  who only has one job, to read these quadrature encoders
```
AdamuKaapan, line 66: And now the system has more than one job, hahahaha: is this comment still relevant?

AdamuKaapan (dismissed):
Looks good to me, just had a question about the mixing of floats and doubles (which is probably a very irrelevant/small comment).

AdamuKaapan (dismissed):

RyanBerge (dismissed):

PR #52, opened 2018-03-25
Fix Test Bug

collinsc:


AdamuKaapan (approved):
We might want to put a note into `readme.md` about how you'll need to re-add these lines if you add tests in these packages (that can be done now or later)

PR #53, opened 2018-03-25
Dumping integration

collinsc:
# Description
Integrates control with dumping
# Testing
Video demo
https://youtu.be/CmBEjvzmKb0

PR #54, opened 2018-03-27
Executive Autonomous Integration

collinsc:
# Prerequisites
#53 
# Description
Integrated the autonomous action server and started to clean up our file system. 

This was a pretty big ordeal, and required me to squash bugs/tweak parameters all over our system. For that I am sorry about the scope and messiness of this pull request, but that is what the work required.

Since I was going all over the system I did my best to get rid of any old irrelevant test code where I could. Though I know I did not get it all, that will continue to trickle in over the course of the week.

I was originally planning to do teleop here as well, but this pull request was just getting too big.

Video Demo
https://www.youtube.com/channel/UCRE81Q_E53j1h__7XorcCGQ/videos?disable_polymer=1


src/tfr_control/src/arm_action_server.cpp
```diff
@@ -81,7 +81,7 @@ class ArmActionServer {
             ROS_DEBUG("Executing movement");
             move_group.asyncExecute(my_plan);
 
-            ros::Rate rate(50);
+            ros::Rate rate(10);
```
RyanBerge, line 84: What is the significance of this change?

src/tfr_control/src/control.cpp
```diff
@@ -10,45 +10,136 @@
...
+// ADAM'S TEST CODE
 // TODO adam delete this test code when ready
 // Whether we're running on hardware or using fake values
 const bool use_fake_values = true;
-class StopManager
+// If we're faking the inputs, we need to know the model constraints on
+// the arm: load them here.
+// If not, just use zeroes, the limits don't matter. TEST code
+double lower_limits[tfr_control::RobotInterface::JOINT_COUNT] = {};
+double upper_limits[tfr_control::RobotInterface::JOINT_COUNT] = {};
```
RyanBerge: Now that these aren't scoped to a class, let's put them in an anonymous namespace so that they're still scoped to this compilation unit.

src/tfr_control/src/robot_interface.cpp
```diff
@@ -182,12 +181,62 @@ namespace tfr_control
         drivebase_v0.second = velocity_values[static_cast<int>(Joint::RIGHT_TREAD)];
```
RyanBerge, line 181: What is the purpose of these `UPKEEP` statements?

src/tfr_control/src/robot_interface.cpp
```diff
@@ -182,12 +181,62 @@ namespace tfr_control
...
+    }
+
+    /*
+     * Returns if the bin is extended or not
+     * */
+    bool RobotInterface::isBinExtended()
+    {
+        double goal = 0.785398;
+        double tolerance = 0.05;
+        return goal - position_values[static_cast<int>(Joint::BIN)] < tolerance;
```
RyanBerge: May want to turn these into constants or node parameters.

src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -45,7 +52,7 @@ class Dumper
         
         Dumper(ros::NodeHandle &node, const std::string &service_name,
                 const DumpingConstraints &c) :
-            server{node, "dumping_action_server", boost::bind(&Dumper::dump, this, _1), false},
+            server{node, "dump", boost::bind(&Dumper::dump, this, _1), false},
             detector{"light_detection"},
             aruco{"aruco_action_server",true},
             //TODO add dumping controller here
```
RyanBerge: Did you intend to fulfill this TODO in this PR?

src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -124,8 +137,22 @@ class Dumper
...
+            bin_cmd.data = BIN_EXTENDED;
+            tfr_msgs::QuerySrv query;
+            ros::Rate rate(10);
+            while (!server.isPreemptRequested() && ros::ok())
+            {
+                ros::service::call("is_bin_extended", query);
+                if (query.response.data)
+                    break;
+                bin_publisher.publish(bin_cmd);
+                stopMoving();
```
RyanBerge: I doubt this would make much of a difference, but we should probably call `stopMoving()` before we do a service call so if something goes slow with the call we aren't just driving into the bumper.

RyanBerge: Also, do you want to call `stopMoving()` in every iteration of the loop while the bin is being raised?

Also also, shouldn't there be a call somewhere to begin raising the bin?

src/tfr_executive/src/autonomous_action_server.cpp
```diff
@@ -80,39 +124,189 @@ class AutonomousExecutive
...
+            }
+
+            if (NAVIGATION_TO)
+            {
+                ROS_INFO("Autonomous Action Server: commencing navigation");
+ 
+                tfr_msgs::NavigationGoal goal;
+                //messages can't support user defined types
+                goal.location_code= static_cast<uint8_t>(tfr_utilities::LocationCode::MINING);
+                navigationClient.sendGoal(goal);
```
RyanBerge, line 171: Can you describe some of the data flow for me here?  Navigation needs to know the point where we're going, and I thought that was calculated during Localization.  If Localization isn't returning that (or maybe more accurately, returning the robot's relative position), how is that information getting to Navigation?

src/tfr_launch/launch/robot.launch
```diff
@@ -0,0 +1,11 @@
...
+<launch>
+    <include file="$(find tfr_launch)/launch/core.launch"/>
+    <include file="$(find tfr_executive)/launch/executive.launch"/>
+    <include file="$(find tfr_sensor)/launch/sensor.launch"/>
+    <include file="$(find tfr_control)/launch/control.launch"/>
+    <include file="$(find tfr_localization)/launch/localization.launch"/>
+    <include file="$(find tfr_navigation)/launch/navigation.launch"/>
+    <include file="$(find tfr_mining)/launch/mining.launch"/>
+    <include file="$(find tfr_dumping)/launch/dumping.launch"/>
+</launch>
```
RyanBerge, line 11: This is an exciting file.

RyanBerge (changes requested):
This looks pretty good, Collin. I have a few questions and a couple comments but it's nothing game-breaking.  Good work.

src/tfr_control/src/arm_action_server.cpp
collinsc: Well we run all our action servers at around 10 hz, which is enough to meet our response requirements. I had to make some modifications and I saw I could free up a few cycles. So relatively benign.

src/tfr_control/src/control.cpp
collinsc: you got it.

src/tfr_control/src/robot_interface.cpp
collinsc: This happens in #57

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: Oh it's fulfilled, I just forgot about the comment.

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: We can move stop moving before the loop. I was just paranoid about any extra messages lingering in the network, and messing up our navigational aid. As far as repeatedly sending stop messages, I could take it or leave it. We are not in performance critical code, but it does look a little sloppy and excessive in the loop. 

As far as raising the bin. We use a position controller, so we make a message with the angle in radians that we want the bin to move to, publish it to the controller's `command` topic and the controller will work with the control layer to take care of the rest.

Most controllers require messages to be resent periodically or else they will consider their command "stale". Once it reaches it's goal, it will just hold position.

Let me know if that clarified what you wanted answered about the control loop.

src/tfr_executive/src/autonomous_action_server.cpp
collinsc: Happy to.
_Localization_ Here we figure out from the cameras where we are. We take that information and send it to a node called the `bin_broadcaster`, it's job is to repeatedly broadcast transforms of where the bin is in the `tf2` transform system, which is integrated with our odometry and sensors. From the moment we can translate any point, from any reference frame in the system and find it's distance from the bin.

_Navigation_ We know we to travel 5.2 m east of the bin. Here is how that works in code The navigation stack takes an xyz quaternion command, in relation to a `tf2` reference frame. All we have to do  is tell it to go +5.2m in the "east" direction, and say that these measurements are in relation to the bin reference frame, and it will take us to the correct place.

collinsc: However what I have just described happens at the navigation action server, and localization action server level. Up here at executive we don't have as much detail.

src/tfr_launch/launch/robot.launch
collinsc: I know right.

src/tfr_control/src/robot_interface.cpp
collinsc: It's for the acceleration limiting. I use the delta_t and delta_v, to take a programmers derivative.

src/tfr_dumping/src/dumping_action_server.cpp
RyanBerge: For some reason I didn't see the `bin_publisher.publish()` call.

src/tfr_executive/src/autonomous_action_server.cpp
RyanBerge: So the transform publisher is set up during Localization and continues to execute during Navigation?

collinsc:
@RyanBerge I think I was able to hit all of your comments and make your requested revisions. Let me know if I missed anything.

RyanBerge (dismissed):
Looks good to me. Very exciting.

src/tfr_executive/src/autonomous_action_server.cpp
```diff
@@ -80,39 +124,189 @@ class AutonomousExecutive
...
         }
 
         actionlib::SimpleActionServer<tfr_msgs::EmptyAction> server;
+
+        actionlib::SimpleActionClient<tfr_msgs::EmptyAction> localizationClient;
+        actionlib::SimpleActionClient<tfr_msgs::NavigationAction> navigationClient;
+        actionlib::SimpleActionClient<tfr_msgs::DiggingAction> diggingClient;
+        actionlib::SimpleActionClient<tfr_msgs::EmptyAction> dumpingClient;
+
+        bool LOCALIZATION;
```
LinuxKid21, line 304: Do we want to make these uppercase like this? It doesn't seem consistent with the TFR coding style.

LinuxKid21 (approved):
A small critique, but nothing that should stop it from getting merged being as time is quite a priority.

PR #55, opened 2018-03-27
PWM Safety Integration

collinsc:
# Prerequisites
#54 
# Description
The hardware designers for the PCA9685 chip have a feature in their chip that isn't desirable in our set of circumstances. In the event of power loss, and restart, when the chip boots it will immediately start broadcasting the same signals it did before the power loss. This means if we ever had to trigger an e stop, or had an unexpected shutdown, the second it got power it would keep travelling in it's last trajectory or move the arm in the last path, before our software could boot and control it. 

This could cause bodily harm or damage to the robot. So the solution was to use the output enable line on the chip to disable output, until expicitly commanded to start by our control system. 

We already coded this feature, this pull request just added the buttons in the gui, and I tested it with a voltmeter. 

Our power system is under construction so we will do tests with our motors when we wire it friday.


RyanBerge (approved):
I don't see anything weird here. Is the code in this PR related to the e-stop problems you were describing the other day? If not, I approve. It yes, let's address that first.

collinsc:
@RyanBerge It wasn't related, after this test we ended up having some hardware problems with the AND gate. John and I tried it with an AND gate and this impelmentation works the way we want it to now.

PR #56, opened 2018-03-29
added board generation and first board design

LinuxKid21:


collinsc (dismissed):
@LinuxKid21 Can you answer what get's reported back from the markers, is it the center of the board?

RyanBerge (dismissed):

RyanBerge (dismissed):

PR #57, opened 2018-03-29
Teleop refinement

collinsc:
# Prerequisites
#54 
# Description
Integrated our system with teleop! 
Video demo: https://youtu.be/Ppc6rWnbfTM

### DIFF
--- a/src/tfr_control/src/robot_interface.cpp
+++ b/src/tfr_control/src/robot_interface.cpp
@@ -169,8 +169,7 @@ namespace tfr_control
         pwm.set(PWMInterface::Address::TREAD_LEFT, 1);
 
         //RIGHT_TREAD
-        signal =
-            drivebaseVelocityToPWM(command_values[static_cast<int>(Joint::RIGHT_TREAD)],
+        signal = drivebaseVelocityToPWM(command_values[static_cast<int>(Joint::RIGHT_TREAD)],
                     drivebase_v0.second);
         pwm.set(PWMInterface::Address::TREAD_RIGHT, signal);
 
@@ -232,7 +231,7 @@ namespace tfr_control
     }
 
     /*
-     * Retrieved the state of the bin
+     * Retrieves the state of the bin
      * */
     double RobotInterface::getBinState()
     {
--- a/src/tfr_control/src/robot_interface.cpp
+++ b/src/tfr_control/src/robot_interface.cpp
@@ -27,9 +27,9 @@ namespace tfr_control
 
         // Connect and register each joint with appropriate interfaces at our
         // layer
-        registerTreadJoint("left_tread_joint", Joint::LEFT_TREAD);
-        registerTreadJoint("right_tread_joint", Joint::RIGHT_TREAD);
-        registerBinJoint("bin_joint", Joint::BIN); 
+        registerJoint("left_tread_joint", Joint::LEFT_TREAD);
+        registerJoint("right_tread_joint", Joint::RIGHT_TREAD);
+        registerJoint("bin_joint", Joint::BIN); 
         registerArmJoint("turntable_joint", Joint::TURNTABLE);
         registerArmJoint("lower_arm_joint", Joint::LOWER_ARM);
         registerArmJoint("upper_arm_joint", Joint::UPPER_ARM);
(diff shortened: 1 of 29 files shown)


src/tfr_control/src/robot_interface.cpp
```diff
@@ -160,7 +165,8 @@ namespace tfr_control
         //LEFT_TREAD
         signal = drivebaseVelocityToPWM(command_values[static_cast<int>(Joint::LEFT_TREAD)],
                     drivebase_v0.first);
-        pwm.set(PWMInterface::Address::TREAD_LEFT, signal);
+        //TODO
+        pwm.set(PWMInterface::Address::TREAD_LEFT, 1);
 
         //RIGHT_TREAD
         signal =
```
RyanBerge: What's with the weird spacing here?

src/tfr_control/src/robot_interface.cpp
```diff
@@ -226,37 +232,31 @@ namespace tfr_control
     }
 
     /*
-     * Returns if the bin is extended or not
+     * Retrieved the state of the bin
```
RyanBerge: Retrieves*

src/tfr_dumping/launch/dumping.launch
```diff
@@ -1,5 +1,9 @@
 <launch>
-    <node name="light_detection_action_server" pkg="tfr_sensor" type="light_detection_action_server"/>
+    <node name="light_detection_action_server" pkg="tfr_sensor" type="light_detection_action_server">
+        <rosparam>
+            threshold: 0.1
```
RyanBerge, line 4: I assume this number represents some difference in light for dumping?

src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -62,7 +63,7 @@ class Dumper
         {
             ROS_INFO("dumping action server initializing");
             detector.waitForServer();
-            aruco.waitForServer();
+            //aruco.waitForServer();
```
RyanBerge: Why is this call commented out?

src/tfr_dumping/src/dumping_action_server.cpp
```diff
@@ -94,13 +93,13 @@ class Dumper
...
-                server.setAborted();
-                return;
-            }
+//            tfr_msgs::ArucoResult initial_estimate{};
+//            getArucoEstimate(initial_estimate);
+//            if (initial_estimate.number_found == 0)
+//            {
+//                server.setAborted();
+//                return;
+//            }
```
RyanBerge: Were these commented out for testing? If so, that change probably shouldn't be merged into the codebase.

src/tfr_utilities/include/tfr_utilities/control_code.h
```diff
@@ -0,0 +1,28 @@
...
+        static const float ARM_TURNTABLE_MIN = 0.0;
+        static const float ARM_LOWER_MAX = 1.69297;
+        static const float ARM_LOWER_MIN = 0.20944;
+        static const float ARM_UPPER_MAX = 1.8675;
+        static const float ARM_UPPER_MIN = 0.0;
+        static const float ARM_SCOOP_MAX = 1.3446;
+        static const float ARM_SCOOP_MIN = -0.872665;
+        static const float BIN_MAX = 0.785398;
+        static const float BIN_MIN = 0.0;
+    }
```
RyanBerge, line 25: I don't suppose there's a way to load the constants into here and the model from one place, or load from the model into here?

RyanBerge (changes requested):
A few questions and comments.

src/tfr_control/src/robot_interface.cpp
collinsc: My editor autowraps sometimes without me noticing.

src/tfr_dumping/launch/dumping.launch
collinsc: Yes it does, but  in this pr it doesn't work right yet. It's final form is in https://github.com/TrickfireRobotics/NasaRmc2018/pull/60

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: you are right that needs to be cleaned up, I probably just forgot

src/tfr_utilities/include/tfr_utilities/control_code.h
collinsc: I had the same thought, and it was not clean.

src/tfr_dumping/src/dumping_action_server.cpp
collinsc: I remember why it's because of levi's aruco bug

src/tfr_utilities/include/tfr_utilities/control_code.h
RyanBerge: I just feel like it's a bug waiting to happen. We get the arm onto the robot and refine some of the constants to match real life but then they don't get changed in both places...

RyanBerge (approved):
Thanks for making those changes. I've approved.

src/tfr_utilities/include/tfr_utilities/control_code.h
collinsc: It could be, but the overal geometry of the arm will not change frequently that would require a cutting torch and a very grumpy EE.

collinsc: I can make a maintenance card for it, if we have some time and or figure out a better way to go about it during the arm we can do that.

PR #58, opened 2018-04-02
Motor integration

collinsc:
Added second arduino in preparation for wednesday.

### DIFF
--- a/src/tfr_control/src/robot_interface.cpp
+++ b/src/tfr_control/src/robot_interface.cpp
@@ -16,7 +16,10 @@ namespace tfr_control
     RobotInterface::RobotInterface(ros::NodeHandle &n, bool fakes, 
             const double *lower_lim, const double *upper_lim) :
         pwm{},
-        arduino{n.subscribe("/sensors/arduino", 5, &RobotInterface::readArduino, this)},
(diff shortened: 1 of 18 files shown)


src/tfr_control/src/robot_interface.cpp
```diff
@@ -118,9 +124,10 @@ namespace tfr_control
...
-        tfr_msgs::ArduinoReading reading;
-        if (latest_arduino != nullptr)
-            reading = *latest_arduino;
+        tfr_msgs::ArduinoAReading reading_a;
+        tfr_msgs::ArduinoBReading reading_b;
+        if (latest_arduino_a != nullptr)
+            reading_a = *latest_arduino_a;
 
         double signal;
         if (use_fake_values) //test code  for working with rviz simulator
```
RyanBerge, line 133: I'm looking forward to the day where we can finally delete this `if` clause here.

RyanBerge (approved):
Looks good to me.

src/tfr_control/src/robot_interface.cpp
collinsc: Lol I think we all are.

PR #61, opened 2018-04-08
Update Network Initialization Scripts

Jebbs:
This includes the stuff that actually starts each system on both Mission Control and the robot.


RyanBerge (approved):
Looks good.

PR #64, opened 2018-04-14
Aruco fix

collinsc:
Fixed the aruco sheets, fixed and debugged the localization system.


RyanBerge (approved):
At some point I would like to refactor the `localization_action_server` and `fiducial_odom_publisher` to use a single class, since they're basically the same code, but that can wait.

collinsc:
I'll make a maintenence trello card for it

PR #65, opened 2018-04-14
Motor demo

collinsc:
PID tuning for left and right tread, added pwm delta limit to fix brownout problem

### DIFF
--- a/src/tfr_control/src/robot_interface.cpp
+++ b/src/tfr_control/src/robot_interface.cpp
@@ -71,7 +71,7 @@ namespace tfr_control
 
         //LEFT_TREAD
         position_values[static_cast<int>(Joint::LEFT_TREAD)] = 0;
-        velocity_values[static_cast<int>(Joint::LEFT_TREAD)] = reading_a.tread_left_vel;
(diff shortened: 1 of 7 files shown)


src/tfr_control/src/robot_interface.cpp
```diff
@@ -71,7 +71,7 @@ namespace tfr_control
 
         //LEFT_TREAD
         position_values[static_cast<int>(Joint::LEFT_TREAD)] = 0;
-        velocity_values[static_cast<int>(Joint::LEFT_TREAD)] = reading_a.tread_left_vel;
+        velocity_values[static_cast<int>(Joint::LEFT_TREAD)] = -reading_a.tread_left_vel;
```
RyanBerge, line 74: This seems like an important change...

Why would this be required?

RyanBerge (approved):

src/tfr_control/src/robot_interface.cpp
collinsc: Counterclockwise is positive, clockwise is negative, to make that match up with our right angle gearboxes you flip them

PR #67, opened 2018-04-23
updated aruco for new board

LinuxKid21:
and reduced the refinement size from 10 to 5 as it was too aggressive for the small markers.


collinsc (approved):
Been using this with no bugs for a week

